PR #8, opened 2022-05-10
Seabeta 197 bre api integration

natdeanlewissoftwire:
This PR integrates the BRE API by building a request from the user's answers on the site and then displaying the measures returned by the BRE API.

Tested: Locally by walking through a range of user stories
Risks: If BRE API returns unexpected result that could cause failure, mitigated by expected measures being specified in the request.
Documentation: BRE API details added to keeper with instructions, comments added in code where assumptions have been made to fit questions asked with BRE API inputs - these values may need to be updated later if the questions asked of the user change.


SeaPublicWebsite/ExternalServices/Models/RequestEpc.cs
```diff
@@ -0,0 +1,43 @@
+﻿namespace SeaPublicWebsite.ExternalServices.Models
```
DanCorderSoftwire, line 1: I don't think this is used anywhere? If that's correct please delete it

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1154,11 +1155,12 @@ public IActionResult YourRecommendations_Get(string reference)
             ).ToList();
             userDataStore.SaveUserData(userDataModel);
```
DanCorderSoftwire, line 1156: In the code above here we take a `Recommendation` object and turn it into a `UserRecommendation` object with exactly the same fields. Can we scrap one of those classes?

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -1,130 +1,433 @@
...
 
-            // Glazing is single or both
-            if (userData.GlazingType is GlazingType.Both or GlazingType.SingleGlazed)
+            return request;
+        }
+
+        private static int GetConvertedPropertyType(PropertyType? propertyType)
+        {
+            //ApartmentFlatOrMaisonette is assumed to be a Flat
+            if (propertyType != null) return (int) (PropertyTypeEnum) propertyType;
```
DanCorderSoftwire, line 246: This is relying on the two enums having the same int values for the same meaning, but there are no comments on the enums to warn anyone that changing them will break this mapping.
This code should really do an explicit switch statement so that the underlying values don't matter

DanCorderSoftwire, line 246: Please always use `{}` with `if` statements

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -1,130 +1,433 @@
...
+                    "U"
+                }
             };
 
-            // Glazing is single or both
-            if (userData.GlazingType is GlazingType.Both or GlazingType.SingleGlazed)
+            return request;
+        }
+
+        private static int GetConvertedPropertyType(PropertyType? propertyType)
```
DanCorderSoftwire, line 243: Why return an `int` when you have a `PropertyTypeEnum` value?

SeaPublicWebsite/ExternalServices/Models/PropertyTypeEnum.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace SeaPublicWebsite.ExternalServices.Models
```
DanCorderSoftwire, line 1: This looks to be specifically for the BRE API with values that match the API spec.
If that is correct then I think this should be called something like `BrePropertyType`, the values should be explicitly set, and there should be a comment explaining that the values must match the BRE API spec.
This probably applies to all the new enums

SeaPublicWebsite/Services/RecommendationService.cs
DanCorderSoftwire, line 243: For all these methods I think it would make sense to swap `Converted` with `Bre`

SeaPublicWebsite/ExternalServices/Models/BreRequest.cs
```diff
@@ -0,0 +1,39 @@
+﻿namespace SeaPublicWebsite.ExternalServices.Models
```
DanCorderSoftwire, line 1: This file and all the BRE specific enums and the API file would be better off in a specific folder together like `ExternalServices/Bre/`

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -1,130 +1,433 @@
...
+                    break;
+                case PropertyType.ApartmentFlatOrMaisonette:
+                    flatLevel = flatType switch
+                    {
+                        FlatType.TopFloor => (int) FlatLevelEnum.TopFloor,
+                        FlatType.MiddleFloor => (int) FlatLevelEnum.MidFloor,
+                        FlatType.GroundFloor => (int) FlatLevelEnum.GroundFloor,
+                        _ => throw new ArgumentOutOfRangeException()
+                    };
+                    //assumption:
```
DanCorderSoftwire, line 289: I think this needs a better comment e.g. `builtForm doesn't make sense for a flat, but is currently a mandatory field on the BRE API so we set MidTerrace as a default`

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -1,130 +1,433 @@
...
+                        HouseType.Terraced => (int) BuiltFormEnum.MidTerrace,
+                        _ => throw new ArgumentOutOfRangeException()
+                    };
+                    break;
+                case PropertyType.Bungalow:
+                    builtForm = bungalowType switch
+                    {
+                        BungalowType.Detached => (int) BuiltFormEnum.Detached,
+                        BungalowType.SemiDetached => (int) BuiltFormEnum.SemiDetached,
+                        //assumption:
```
DanCorderSoftwire, line 274: Why is this an assumption?

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -1,130 +1,433 @@
...
+        }
+
+        private static int GetConvertedPropertyType(PropertyType? propertyType)
+        {
+            //ApartmentFlatOrMaisonette is assumed to be a Flat
+            if (propertyType != null) return (int) (PropertyTypeEnum) propertyType;
+            throw new ArgumentNullException();
+        }
+
+        private static (int builtForm, int? flatLevel) GetConvertedBuiltFormAndFlatLevel(PropertyType? propertyType,
```
DanCorderSoftwire, line 250: Why return `int`s instead of the enum values?

DanCorderSoftwire, line 250: I'm not keen on returning multiple values from a function if we don't have to. I'm not sure that this gives us anything better than having two separate functions would
`private static BreBuiltForm GetBreBuiltForm(propertyType)`
`private static BreFlatLevel? GetBreFlatLevel(propertyType)`

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -1,130 +1,433 @@
...
+                WallConstruction.Cavity => cavityWallsInsulated switch
+                {
+                    CavityWallsInsulated.DoNotKnow => (int) WallTypeEnum.DontKnow,
+                    CavityWallsInsulated.No => (int) WallTypeEnum.CavityWallsWithoutInsulation,
+                    CavityWallsInsulated.Some => (int) WallTypeEnum.CavityWallsWithoutInsulation,
+                    CavityWallsInsulated.All => (int) WallTypeEnum.CavityWallsWithInsulation,
+                    _ => throw new ArgumentOutOfRangeException()
+                },
+                WallConstruction.Mixed =>
+                    //assumption:
```
DanCorderSoftwire, line 343: How are we going to make sure that we come back to all of these assumptions? Are they listed on the ticket?

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -1,130 +1,433 @@
...
-                && !UserDataHelper.HasInsulatedFloor(userData))
+        private static int GetConvertedRoofType(RoofConstruction? roofConstruction, RoofInsulated? roofInsulated)
+        {
+            return roofConstruction switch
             {
-                userRecommendationKeys.Add(RecommendationKey.GroundFloorInsulation);
-            }
+                RoofConstruction.Flat =>
+                    //assumption:
+                    (int) RoofTypeEnum.DontKnow,
```
DanCorderSoftwire, line 356: Not your fault, but this seems really weird. Why are we asking about flat roofs if we can't do anything with the information?

DanCorderSoftwire (changes requested):

SeaPublicWebsite/ExternalServices/Models/HeatingFuelEnum.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace SeaPublicWebsite.ExternalServices.Models
+{
+    public enum HeatingFuelEnum
+    {
+        SolidFuel = 9,
```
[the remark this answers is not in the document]
natdeanlewissoftwire: Yep, these are the values required by the BRE Api, not sure why they are what they are

SeaPublicWebsite/Helpers/Global.cs
```diff
@@ -18,5 +18,8 @@ public static class Global
         public static string EpbEpcAuthUsername => Config.GetAppSetting("EpbEpcAuthUsername");
         public static string EpbEpcAuthPassword => Config.GetAppSetting("EpbEpcAuthPassword");
         public static string GovUkNotifyApiKey => Config.GetAppSetting("GovUkNotifyApiKey");
+        public static string BreUsername => Config.GetAppSetting("BreUsername");
+        public static string BrePassword => Config.GetAppSetting("BrePassword");
+        public static string BreBaseAddress => Config.GetAppSetting("BreBaseAddress");
```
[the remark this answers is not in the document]
natdeanlewissoftwire: ah thanks, I'd had it in my appsettings.secret.json instead

SeaPublicWebsite/ExternalServices/BreApi.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+            catch (Exception)
+            {
+                return null;
+            }
+        }
+
+        private static string GenerateNonce()
+        {
+            string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
```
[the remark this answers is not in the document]
natdeanlewissoftwire: yep, have checked with Debs and she's confirmed the nonce is not sent from the server

SeaPublicWebsite/ExternalServices/BreApi.cs
```diff
@@ -0,0 +1,95 @@
...
+                            recommendations.Add(recommendation);
+                        }
+                        return recommendations;
+                    }
+
+                    return null;
+                }
+            }
+            catch (Exception)
+            {
```
DanCorderSoftwire, line 80: Please add a todo for seabeta-192 to add a log here

SeaPublicWebsite/ExternalServices/BreApi.cs
```diff
@@ -0,0 +1,95 @@
...
+                                LifetimeSaving = lifetime * saving,
+                                Lifetime = lifetime,
+                                Summary = recommendationDictionary[prop.Name].Summary
+                            };
+                            recommendations.Add(recommendation);
+                        }
+                        return recommendations;
+                    }
+
+                    return null;
```
DanCorderSoftwire, line 76: This null looks like it will blow up into a null ref exception further up the stack. We should handle BRE API failures more explicitly. Potentially by throwing an exception here so the user sees a more helpful error message (we don't have a nice error page yet though).
Same for catching an exception below.

SeaPublicWebsite/ExternalServices/BreApi.cs
```diff
@@ -0,0 +1,95 @@
...
+                        (DateTimeFormatInfo.InvariantInfo.SortableDateTimePattern) + "Z";
+                    string token = GenerateToken(password + nonce + username + created);
+                    string wsseHeader =
+                        $"WSSE UsernameToken Token=\"{token}\", Nonce=\"{nonce}\", Username=\"{username}\", Created=\"{created}\"";
+                    httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", wsseHeader);
+
+                    string path = "/bemapi/energy_use";
+                    string requestString = JsonConvert.SerializeObject(request);
+                    StringContent stringContent = new(requestString);
+                    HttpResponseMessage response = await Task.FromResult(httpClient.PostAsync(path, stringContent)).Result;
```
DanCorderSoftwire, line 42: This should just be `HttpResponseMessage response = await httpClient.PostAsync(path, stringContent);`
Same below

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -180,49 +181,48 @@ public static List<Recommendation> GetRecommendationsForUser(UserDataModel userD
         {
             BreRequest request = CreateRequest(userData);
 
-            string requestString = JsonConvert.SerializeObject(request);
-            return BreApi.GetRecommendationsForUserRequest(requestString);
+            return BreApi.GetRecommendationsForUserRequest(request).Result;
```
DanCorderSoftwire, line 184: Please don't use `.Result` with async method unless you really understand async stuff. This is a potential deadlock again.
Once a method becomes async then generally everything above it in the call stack needs to become async too (which can be a pain).

SeaPublicWebsite/ExternalServices/BreApi.cs
```diff
@@ -6,16 +6,19 @@
...
+using SeaPublicWebsite.ExternalServices.Models;
 using SeaPublicWebsite.Helpers;
 using SeaPublicWebsite.Services;
 
 namespace SeaPublicWebsite.ExternalServices
 {
     public static class BreApi
     {
-        public static List<Recommendation> GetRecommendationsForUserRequest(string requestString)
+        public static async Task<List<Recommendation>> GetRecommendationsForUserRequest(BreRequest request)
```
DanCorderSoftwire, line 21: Please add `Async` to the end of this function name (it's a good idea for pretty much any async function so people using it know to `await` it)

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Services/RecommendationService.cs
natdeanlewissoftwire: Ah okay, I've awaited (and suffixed with Async) the following: BreApi.GetRecommendationsForUserRequestAsync, RecommendationService.GetRecomendationForUserAsync, EnergyEfficiencyController.YourRecommendations_GetAsync.

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
DanCorderSoftwire: Ok, fair enough. Maybe rename `Recommendation` to `BreRecommendation` and move it into the Bre folder though so it's more clear what the distinction is.

SeaPublicWebsite/Services/RecommendationService.cs
DanCorderSoftwire: Could you raise this with Dharmesh please, just so it doesn't get forgotten about.

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -1,130 +1,438 @@
...
+            bool? breHotWaterCylinder = GetBreHotWaterCylinder(userData.HasHotWaterCylinder);
+
+            BreHeatingPatternType breHeatingPatternType = GetBreHeatingPatternType(userData.HeatingPattern);
+
+            BreRequest request = new()
             {
-                // Always shown
-                RecommendationKey.UpgradeHeatingControls
+                postcode = userData.Postcode,
+                property_type = ((int) brePropertyType).ToString(),
```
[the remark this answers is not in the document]
DanCorderSoftwire: That's unfortunate :(
As we discussed, please move this horribleness into a BreRequest constructor so we don't have to look at it so often :)

SeaPublicWebsite/ExternalServices/BreApi.cs
```diff
@@ -73,11 +73,12 @@ public static async Task<List<Recommendation>> GetRecommendationsForUserRequest(
                         return recommendations;
                     }
 
-                    return null;
+                    throw new ArgumentNullException();
```
DanCorderSoftwire, line 76: This isn't really what `ArgumentNullException`s are for. They are meant to be used when a function is given `null` when it shouldn't be. (Also this is still returning null below in the `catch` block)
Given that we're not sure how we'll be handling errors yet maybe just go with `throw new Exception($"BRE API returned an error: {response.StatusCode}")` for now (and something similar below)

DanCorderSoftwire (changes requested):

SeaPublicWebsite/package-lock.json
```diff
@@ -1,244 +1,8 @@
 {
   "name": "simpleenergyadvicepublicwebsite",
   "version": "1.0.0",
```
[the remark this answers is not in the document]
natdeanlewissoftwire: Ah my mistake - this doesn't look like it's in the net PR though just in the commit history, should I really delete it from the history or leave it as is?

PR #13, opened 2022-05-11
Seabeta 256 delete welcome page

CodrinCotarlanSoftwire:


SeaPublicWebsite/SeaPublicWebsite.csproj
```diff
@@ -31,4 +31,8 @@
         </None>
     </ItemGroup>
 
+    <ItemGroup>
```
DanCorderSoftwire, line 34: I don't think you want this

SeaPublicWebsite/Views/Shared/_Layout.cshtml
```diff
@@ -52,8 +52,8 @@
         {
             ServiceName = "Simple Energy Advice service",
             AssetsPath = assetsPath,
-            HomepageUrl = "/",
-            ServiceUrl = "/",
+            HomepageUrl = "https://www.gov.uk",
```
DanCorderSoftwire, line 55: Does this break the link on local sites? Should it go in appsettings so we can override it?

SeaPublicWebsite/SeaPublicWebsite.csproj
CodrinCotarlanSoftwire: Huh, I wonder why it got added here... yeah, it's not needed I'll remove it.

SeaPublicWebsite/Views/Shared/_Layout.cshtml
CodrinCotarlanSoftwire: HomepageUrl is the link that the left part of the banner (crown GOV.UK) redirects to. For what it's worth, the gender pay gap project did it similarly.
ServiceUrl is the link that the text of the banner redirects to. We can either make it redirect to /energy-efficiency, or to '/' since the home controller will redirect it anyway. What do you think?

DanCorderSoftwire: Ah Ok, I was assuming that `HomepageUrl` would be our service's homepage

DanCorderSoftwire (approved):

PR #15, opened 2022-05-12
Seabeta 261 return to first reccomentation

CodrinCotarlanSoftwire:
I've added a redirect to the first recommendation of the user and works as intended. In the future we may want a separate controller for the recommendations, in which case this will be changed anyway.

### DIFF
--- a/SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
+++ b/SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
@@ -57,8 +57,11 @@ namespace SeaPublicWebsite.Controllers
                     ModelState.AddModelError(nameof(NewOrReturningUserViewModel.Reference), "Check you have typed the reference correctly. Reference must be 8 characters.");
                     return View("NewOrReturningUser", viewModel);
                 }
+
+                var userData = userDataStore.LoadUserData(viewModel.Reference);
+                var recommendations = RecommendationService.GetRecommendationsForUser(userData);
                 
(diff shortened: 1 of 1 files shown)


SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -57,8 +57,11 @@ public IActionResult NewOrReturningUser_Post(NewOrReturningUserViewModel viewMod
                     ModelState.AddModelError(nameof(NewOrReturningUserViewModel.Reference), "Check you have typed the reference correctly. Reference must be 8 characters.");
                     return View("NewOrReturningUser", viewModel);
                 }
+
+                var userData = userDataStore.LoadUserData(viewModel.Reference);
+                var recommendations = RecommendationService.GetRecommendationsForUser(userData);
                 
-                return RedirectToAction("YourSavedRecommendations_Get", "EnergyEfficiency", new { reference = viewModel.Reference });
+                return RedirectToAction("Recommendation_Get", "EnergyEfficiency", new { reference = viewModel.Reference, id = (int) recommendations[0].Key });
```
DanCorderSoftwire, line 64: What if the user doesn't have any recommendations?
I've left the same question on the issue.

CodrinCotarlanSoftwire: This is what it was originally there :) From how the code was originally written, it is guaranteed to show a hardcoded suggestion, but the BRE integration doesn't enforce this anymore...

PR #16, opened 2022-05-16
Seabeta 286 cookies

CodrinCotarlanSoftwire:
No analytics functionality, just infra to present the banner and save cookie settings.


SeaPublicWebsite/Controllers/CookieController.cs
```diff
@@ -0,0 +1,78 @@
...
+    public IActionResult CookieSettings_Get(bool changesHaveBeenSaved = false)
+    {
+        cookieService.TryGetCookie<CookieSettings>(Request, cookieService.Configuration.CookieSettingsCookieName, out var cookie);
+        
+        var viewModel = new CookieSettingsViewModel
+        {
+            GoogleAnalytics = cookie?.GoogleAnalytics is true,
+            ChangesHaveBeenSaved = changesHaveBeenSaved
+        };
+        TempData["BannerState"] = BannerState.Hide;
```
CodrinCotarlanSoftwire, line 33: See below

SeaPublicWebsite/Controllers/CookieController.cs
```diff
@@ -0,0 +1,78 @@
...
+        return CookieSettings_Get(changesHaveBeenSaved: true);
+    }
+
+    [HttpPost("/cookie-consent")]
+    [ValidateAntiForgeryToken]
+    public IActionResult CookieConsent(CookieConsent cookieConsent)
+    {
+        if (cookieConsent.Consent == "hide")
+        {
+            TempData["BannerState"] = BannerState.Hide;
```
CodrinCotarlanSoftwire, line 56: Because we are redirecting, we cannot store the banner state in a ViewBag since it would get reset. TempData is preserved through the chain of redirects.

SeaPublicWebsite/Helpers/Global.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        Configuration = options.Value;
+    }
+}
+
+public class GlobalConstantsConfiguration
+{
+    public const string ConfigSection = "GlobalConstants";
+    public string ServiceName { get; set; }
+}
```
CodrinCotarlanSoftwire, line 18: Let me know of you think of a better way to add a 'global' type of configuration

SeaPublicWebsite/Startup.cs
```diff
@@ -48,6 +52,21 @@ public void ConfigureServices(IServiceCollection services)
...
         }
+        
+        private void ConfigureGlobalConstants(IServiceCollection services)
+        {
+            services.Configure<GlobalConstantsConfiguration>(
+                configuration.GetSection(GlobalConstantsConfiguration.ConfigSection));
+            services.AddSingleton<GlobalConstants>();
+        }
+
+
```
DanCorderSoftwire, line 63: Extra blank line

SeaPublicWebsite/Startup.cs
```diff
@@ -48,6 +52,21 @@ public void ConfigureServices(IServiceCollection services)
...
+        private void ConfigureCookieService(IServiceCollection services)
+        {
+            services.Configure<CookieServiceConfiguration>(
+                configuration.GetSection(CookieServiceConfiguration.ConfigSection));
+            services.AddSingleton<CookieService>();
```
DanCorderSoftwire, line 68: Why `Singleton`?

SeaPublicWebsite/Controllers/CookieController.cs
DanCorderSoftwire: Does this work at the moment? I believe that `TempData` relies on there being a `session` which we don't currently have (and would like to avoid if possible)

DanCorderSoftwire: I suspect that this might look like it's working because of the logic in `_Layout.cshtml`

SeaPublicWebsite/Helpers/Global.cs
DanCorderSoftwire: I think that, given that this will be for things that are actually constant, we can go much simpler and just a have a `Constants.cs` file in the root of the project along the lines of:

```
public static class Constants
{
    public const string SERVICE_NAME = "Improve your property's energy efficiency"
}
```

SeaPublicWebsite/Services/Cookies/CookieService.cs
```diff
@@ -0,0 +1,58 @@
...
+            catch (JsonException)
+            {
+                // In case of failure, return false as if there was no cookie
+            }
+        }
+
+        cookie = default;
+        return false;
+    }
+    public bool HasSeenCookieLatestMessage(HttpRequest request)
```
DanCorderSoftwire, line 36: blank line please

DanCorderSoftwire, line 36: I find this name a bit confusing, maybe `CookieSettingsAreUpToDate`?

SeaPublicWebsite/Views/Cookie/CookieSettings.cshtml
```diff
@@ -0,0 +1,91 @@
...
+                <p class="govuk-notification-banner__heading">
+                  You’ve set your cookie preferences.
+                </p>
+              </div>
+            </div>
+        }
+
+        <form method="POST" action="@Url.Action("CookieSettings_Post", "Cookie")">
+            @(Html.AntiForgeryToken())
+            <h1 class="govuk-heading-xl">Cookies on GOV.UK</h1>
```
DanCorderSoftwire, line 26: A more general question that may need to be asked to BEIS technical people.
This message makes it sound like we're setting cookies for all of Gov.UK when we're actually only doing it for our service. If all services follow this pattern then users of various Gov.UK services will keep seeing the same cookie banner and may wonder why they keep getting asked the same question.
Is that correct, or are we missing something?

SeaPublicWebsite/Views/EnergyEfficiency/Index.cshtml
```diff
@@ -1,4 +1,5 @@
-﻿@{
+﻿@model SeaPublicWebsite.Models.EnergyEfficiency.IndexViewModel
```
DanCorderSoftwire, line 1: If we go with simple constants then we won't need a view model here

SeaPublicWebsite/Controllers/CookieController.cs
```diff
@@ -0,0 +1,78 @@
...
+            Version = cookieService.Configuration.CurrentCookieMessageVersion,
+            GoogleAnalytics = viewModel.GoogleAnalytics,
+        };
+        cookieService.SetCookie(Response, cookieService.Configuration.CookieSettingsCookieName, cookieSettings);
+        return CookieSettings_Get(changesHaveBeenSaved: true);
+    }
+
+    [HttpPost("/cookie-consent")]
+    [ValidateAntiForgeryToken]
+    public IActionResult CookieConsent(CookieConsent cookieConsent)
```
DanCorderSoftwire, line 52: Can you please check that all view model classes have `ViewModel` at the end of their names?

SeaPublicWebsite/Services/Cookies/CookieService.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+        return false;
+    }
+
+    public bool HasAcceptedGoogleAnalytics(HttpRequest request)
+    {
+        return TryGetCookie<CookieSettings>(request, Configuration.CookieSettingsCookieName, out var cookie) && cookie.GoogleAnalytics;
+    }
+
+    public void SetCookie<T>(HttpResponse response, string cookieName, T cookie, int daysUntilExpiry = 365)
```
DanCorderSoftwire, line 50: If we want `DaysUntilExpiry`to be in the config, then we need to remove the `= 365` from here.

SeaPublicWebsite/Controllers/CookieController.cs
DanCorderSoftwire: If I'm correct, then please replace the references to `TempData` with `ViewBag`

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Controllers/CookieController.cs
CodrinCotarlanSoftwire: ViewBag still doesn't work because we would redirect. However I just got an idea, instead of storing whether the user saw the banner or not in the cookie, we can store what banner we should display. I'll see if I can get that working

SeaPublicWebsite/Startup.cs
CodrinCotarlanSoftwire: Sorry, missed this

SeaPublicWebsite/Controllers/CookieController.cs
CodrinCotarlanSoftwire: Done! All I found was one file name typo.

SeaPublicWebsite/Views/Cookie/CookieSettings.cshtml
CodrinCotarlanSoftwire: We can use 'our service', to avoid confusion and to keep it short.

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Views/Shared/_Layout.cshtml
```diff
@@ -9,13 +9,27 @@
...
+    {
+        bannerState = cookie.BannerState;
+        
+        // If the banner shows the confirmation message, update the cookie so it won't be shown again.
+        if (bannerState is BannerState.ShowAccepted or BannerState.ShowRejected)
+        {
+            cookie.BannerState = BannerState.Hide;
+            CookieService.SetCookie(Context.Response, CookieService.Configuration.CookieSettingsCookieName, cookie);
+        }
+    }
```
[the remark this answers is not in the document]
DanCorderSoftwire: To be clear, by "this logic" I mean working out `bannerState` and potentially updating the cookie value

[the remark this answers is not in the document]
CodrinCotarlanSoftwire: Yes I agree. It was a bit more difficult to get the design of this right, and the final draft of this will also support testing!

SeaPublicWebsite/Controllers/CookieController.cs
```diff
@@ -34,6 +34,7 @@ public IActionResult CookieSettings_Post(CookieSettingsViewModel viewModel)
         var cookieSettings = new CookieSettings
         {
             Version = cookieService.Configuration.CurrentCookieMessageVersion,
+            ConfirmationShown = true,
```
DanCorderSoftwire, line 37: Should this be false? Just because they are on the confirmation page doesn't mean that they have seen the confirmation banner?

SeaPublicWebsite.UnitTests/SeaPublicWebsite.UnitTests.csproj
```diff
@@ -10,6 +10,7 @@
     </PropertyGroup>
 
     <ItemGroup>
+        <PackageReference Include="CabinetOffice.GovUkDesignSystem" Version="1.0.0-d899eac" />
```
DanCorderSoftwire, line 13: Do we really need this here? The tests shouldn't need this library

SeaPublicWebsite.UnitTests/CookieServiceTests.cs
```diff
@@ -0,0 +1,247 @@
...
+    {
+        response.Headers.SetCookie.ToString().Should().Contain($"{key}={ConvertObjectToHttpHeaderSrting(value)}");
+    }
+
+    private string ConvertObjectToHttpHeaderSrting(object o)
+    {
+        return Uri.EscapeDataString(JsonConvert.SerializeObject(o));
+    }
+
+    private CookieSettings[] SetUpCookieSettings()
```
DanCorderSoftwire, line 206: This could just be
```
[DatapointSource] 
    private CookieSettings[] CookieSettings = {
        new CookieSettings // Accepted latest cookies
        {
            Version = LatestVersion,
            ConfirmationShown = true,
            GoogleAnalytics = true
        },
        new CookieSettings // Outdated version
        {
            Version = LatestVersion - 1,
            ConfirmationShown = true,
            GoogleAnalytics = true
        },
        ...
};
```

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Controllers/CookieController.cs
CodrinCotarlanSoftwire: If they reach this it means they submitted their cookie preferences, which means they consented, so we don't need a banner anymore.

SeaPublicWebsite.UnitTests/SeaPublicWebsite.UnitTests.csproj
CodrinCotarlanSoftwire: I believe Rider added it for me, I'll remove it

SeaPublicWebsite/Controllers/CookieController.cs
DanCorderSoftwire: We still need to tell them that they succeeded in setting their preference. However if you save the settings on this page there is a different message at the top of the page that says "You’ve set your cookie preferences" so we don't need the banner, so this is OK.

CodrinCotarlanSoftwire: Yup! I've already added a confirmation message whenever an user saves their settings here.

SeaPublicWebsite/Controllers/CookieController.cs
```diff
@@ -0,0 +1,69 @@
...
+public class CookieController: Controller
+{
+    private readonly CookieService cookieService;
+
+    public CookieController(CookieService cookieService)
+    {
+        this.cookieService = cookieService;
+    }
+
+    [HttpGet("/cookies")]
```
DanCorderSoftwire, line 17: The `/` at the beginning here means that we may look like we're setting cookies for all of GOV.UK

SeaPublicWebsite.UnitTests/CookieServiceTests.cs
CodrinCotarlanSoftwire: More than this, I also needed to create a test case class to add a toString method so that failing cases have a name

SeaPublicWebsite/Controllers/CookieController.cs
```diff
@@ -0,0 +1,69 @@
...
+        var viewModel = new CookieSettingsViewModel
+        {
+            GoogleAnalytics = cookie?.GoogleAnalytics is true,
+            ChangesHaveBeenSaved = changesHaveBeenSaved
+        };
+        return View("CookieSettings", viewModel);
+    }
+
+    [HttpPost("cookies")]
+    [ValidateAntiForgeryToken]
```
DanCorderSoftwire, line 31: Instead of this, could you please add `options.Filters.Add(new AutoValidateAntiforgeryTokenAttribute());` to `services.AddControllersWithViews(options =>` in `Startup.cs`
That way all forms will be automatically validated

SeaPublicWebsite/Views/Cookie/CookieSettings.cshtml
```diff
@@ -0,0 +1,91 @@
...
+              <div class="govuk-notification-banner__content">
+                <p class="govuk-notification-banner__heading">
+                  You’ve set your cookie preferences.
+                </p>
+              </div>
+            </div>
+        }
+
+        <form method="POST" action="@Url.Action("CookieSettings_Post", "Cookie")">
+            @(Html.AntiForgeryToken())
```
DanCorderSoftwire, line 25: This shouldn't be necessary in .net core 6. Please remove this line and check that the token is still generated on the page.

DanCorderSoftwire (changes requested):

PR #17, opened 2022-05-16
Seabeta 275 automated unit testing setup

natdeanlewissoftwire:
Tests - the test passes, it only tests whether the BreRequest constructor requests measures in the response from the API for now. This ticket is just setup, so does not contain any substantial tests.

Risks - minimal, this content is new and in a separate project to the rest of the codebase

No documentation required.


SimpleEnergyAdvice.sln
```diff
@@ -26,6 +26,8 @@ ProjectSection(SolutionItems) = preProject
 	.github\workflows\build-and-deploy-to-gov-paas-usertesting.yml = .github\workflows\build-and-deploy-to-gov-paas-usertesting.yml
 EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Tests", "Tests\Tests.csproj", "{BC27E00A-BCB1-40DD-872A-3FF0A46A5FFF}"
```
DanCorderSoftwire, line 29: Please call this project "SeaPublicWebsite.UnitTests.csproj" thanks

Tests/BreApi.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    [SetUp]
+    public void Setup()
+    {
+    }
+
+    [Test]
+    public void RequestsMeasures()
+    {
+        BreRequest request = new(
```
DanCorderSoftwire, line 17: Please add `Arrange`, `Act`, `Assert` comments to tests so it's clear what's going on where.
In this case `Arrange` and `Act` are the same step (which is fine)

DanCorderSoftwire (changes requested):

PR #18, opened 2022-05-17
Govukdesignsystem dependency

DanCorderSoftwire:
Switch from having a local copy of GovUkDesignSystem to pulling it in as a Nuget package.
Currently a local package as the Cabinet Office don't publish the GovUkDesignSystem to Nuget, but when they do this should be easy to update.


SeaPublicWebsite/SeaPublicWebsite.csproj
```diff
@@ -8,15 +8,12 @@
 
     <ItemGroup>
       <PackageReference Include="AWSSDK.S3" Version="3.7.7.21" />
+      <PackageReference Include="CabinetOffice.GovUkDesignSystem" Version="1.0.0-5981d4a" />
```
CodrinCotarlanSoftwire, line 11: Can the version number look nicer in the future? (Just a mild observation)

CodrinCotarlanSoftwire: Ah then leaving it there sounds good!

README.md
```diff
@@ -10,6 +10,24 @@
 
 In SeaPublicWebsite run `npm install`
 
+### GovUkDesignSystem
+
+We are using the GovUkDesignSystem library from the Cabinet Office: https://github.com/cabinetoffice/govuk-design-system-dotnet
+
+As this library is not currently published to Nuget we have a copy of the library in a nuget package in the /lib folder of this solution.
```
natdeanlewissoftwire, line 17: /lib should be /Lib for consistency

PR #21, opened 2022-05-19
Seabeta 282 ef setup

natdeanlewissoftwire:
Tested locally, database being created and updated correctly in rider and VS.

Risk that user creates different credentials to those hardcoded in appsettings that will need them to tweak their code, this is highlighted in the readme. If a seadev db already exists this could also cause issues but risk is small.

Documentation has been added in README.md for other devs.


SeaPublicWebsite.Data/User.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace SeaPublicWebsite.Data;
```
DanCorderSoftwire, line 1: Can we make this `UserData` rather than `User` please

SeaPublicWebsite.Data/SeaPublicWebsite.Data.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net6.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
```
DanCorderSoftwire, line 6: As we're not using nullable reference types in the main project I think it's going to be confusing to use them just in the data project - especially as EF doesn't seem to play very nicely with nullable reference types anyway: https://docs.microsoft.com/en-us/ef/core/miscellaneous/nullable-reference-types
This will also get rid of the build warning on `SeaDbContext.cs`

DanCorderSoftwire (changes requested):

SeaPublicWebsite.Data/User.cs
natdeanlewissoftwire: Done - i've left the dbset item in dbcontext named user as that still seemed the most sensible choice

SeaPublicWebsite.Data/SeaPublicWebsite.Data.csproj
natdeanlewissoftwire: Done :)

DanCorderSoftwire (approved):

PR #24, opened 2022-05-23
Seabeta 323 answer summary list component

CodrinCotarlanSoftwire:
Branch created before merging 273 (back and forward button refactor), so all 'change =' places will be replaced with 'entryPoint =' after merging.
Also the git diff is not too nice, since I've moved all the body of the page in the 'rows' variable.

### DIFF
--- a/SeaPublicWebsite/Views/EnergyEfficiency/AnswerSummary.cshtml
+++ b/SeaPublicWebsite/Views/EnergyEfficiency/AnswerSummary.cshtml
@@ -20,6 +20,131 @@
     })
 }
 
+@{
+    var rows = new List<SummaryListRowViewModel>();
+    
+    rows.Add(new()
+    {
+        Key = new() 
+        {
+            Classes = "govuk-!-width-one-quarter",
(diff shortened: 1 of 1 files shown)


SeaPublicWebsite/Views/EnergyEfficiency/AnswerSummary.cshtml
```diff
@@ -21,431 +21,672 @@
...
+        {
+            Classes = "govuk-!-width-one-quarter",
+            Text = "Wall type"
+        },
+        Value = new()
+        {
+            Classes = "govuk-!-width-one-half",
+            Html = (@<text>
+                        @(GovUkRadioCheckboxLabelTextAttribute.GetLabelText(Model.WallConstruction))
+                        @if (Model.WallConstruction == WallConstruction.DoNotKnow)
```
DanCorderSoftwire, line 98: Not something for you to fix now, but this logic should not be in a view file :(

DanCorderSoftwire (approved):

PR #30, opened 2022-05-26
Seabeta 349 Add check for valid email address

CodrinCotarlanSoftwire:


SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1143,7 +1144,15 @@ public async Task<IActionResult> YourRecommendations_Post(YourRecommendationsVie
...
             if (viewModel.HasEmailAddress == HasEmailAddress.Yes)
             {
-                emailApi.SendReferenceNumberEmail(userDataModel.EmailAddress, userDataModel.Reference);
+                try
+                {
+                    emailApi.SendReferenceNumberEmail(userDataModel.EmailAddress, userDataModel.Reference);
+                }
+                catch (NotifyClientException)
+                {
+                    ModelState.AddModelError(nameof(viewModel.EmailAddress), "Enter a valid email address");
```
DanCorderSoftwire, line 1153: I suspect that this isn't the only reason that we could get an error from the Notify API so this could be quite annoying to the user if actually they've entered a valid email and the Notify API is down for other reasons.

CodrinCotarlanSoftwire: Hmm, that might be true. I'll add some error messages when it's not the user's fault.

SeaPublicWebsite/ErrorHandling/EmailSenderException.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
```
DanCorderSoftwire, line 1: Instead of `ErrorHandling` can we put this in `EmailSending` please?

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1148,9 +1149,19 @@ public async Task<IActionResult> YourRecommendations_Post(YourRecommendationsVie
...
+                catch (EmailSenderException e)
                 {
-                    ModelState.AddModelError(nameof(viewModel.EmailAddress), "Enter a valid email address");
+                    switch (e.Type)
+                    {
+                        case EmailSenderExceptionType.InvalidEmailAddress:
+                            ModelState.AddModelError(nameof(viewModel.EmailAddress), "Enter a valid email address");
+                            break;
+                        case EmailSenderExceptionType.Other:
+                            ModelState.AddModelError(nameof(viewModel.EmailAddress), "Unable to send email due to unexpected error. Please uncheck this box and make a note of your reference number before you continue.");
```
DanCorderSoftwire, line 1160: We've discussed this, and the error message handling will change a little bit to be more user friendly

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
CodrinCotarlanSoftwire: Preferably the error should be in only one paragraph, so there's no overhead in doing this.

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -191,7 +193,7 @@ public IActionResult AskForPostcode_Get(string reference)
         [HttpPost("postcode/{reference}")]
         public IActionResult AskForPostcode_Post(AskForPostcodeViewModel viewModel)
         {
-            if (!PostcodesIoApi.IsValidPostcode(viewModel.Postcode))
+            if (viewModel.Postcode is not null && !PostcodesIoApi.IsValidPostcode(viewModel.Postcode))
```
CodrinCotarlanSoftwire, line 196: Otherwise we show that an empty postcode is not valid too, when we already have the postcode missing error

PR #33, opened 2022-05-26
Seabeta 347 clone add checkbox

natdeanlewissoftwire:
replace radios with checkbox.

Note: an error is thrown when an email is submitted - this appears to be a notify issue rather than an issue with this ticket, so a fix is not included in this PR:
image


natdeanlewissoftwire:
closed - this change has been added to seabeta-313 instead

PR #38, opened 2022-06-01
Seabeta 345 Add 404 page not found page

CodrinCotarlanSoftwire:


SeaPublicWebsite/Startup.cs
```diff
@@ -116,11 +116,13 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
         {
             if (!webHostEnvironment.IsDevelopment())
             {
-                app.UseExceptionHandler("/Home/Error");
```
CodrinCotarlanSoftwire, line 119: This did nothing

SeaPublicWebsite/Startup.cs
```diff
@@ -116,11 +116,13 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
         {
             if (!webHostEnvironment.IsDevelopment())
             {
-                app.UseExceptionHandler("/Home/Error");
+                app.UseExceptionHandler();
                 // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
                 // app.UseHsts();
             }
 
+            app.UseStatusCodePagesWithReExecute("/error/{0}");
```
CodrinCotarlanSoftwire, line 124: By doing so, on status code errors (i.e. 4xx, 5xx) we redirect to the /error route, which is the ErrorController. There we can handle them any way we want.

SeaPublicWebsite/Views/Error/PageNotFound.cshtml
```diff
@@ -0,0 +1,34 @@
+﻿@{
+    ViewBag.Title = "Page not found - Home Energy Retrofit Advice - GOV.UK";
+}
+
+@section BeforeMain {
```
DanCorderSoftwire, line 5: If this is empty can you just leave it out?

SeaPublicWebsite/Views/Error/ServiceIssue.cshtml
```diff
@@ -0,0 +1,21 @@
+﻿@{
+    ViewBag.Title = "Sorry, there is a problem with the service - Home Energy Retrofit Advice - GOV.UK";
+}
+
+@section BeforeMain {
```
DanCorderSoftwire, line 5: Same here

SeaPublicWebsite/Views/Error/ServiceIssue.cshtml
```diff
@@ -0,0 +1,21 @@
...
+}
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">Sorry, there is a problem with the service</h1>
+        <p class="govuk-body">
+            Try again later.
+        </p>
+        <p class="govuk-body">
+            We have not saved your answers. When the service is available, you will have to start again.
```
DanCorderSoftwire, line 15: This isn't always true?

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Views/Error/PageNotFound.cshtml
CodrinCotarlanSoftwire: Yes i can

SeaPublicWebsite/Views/Error/ServiceIssue.cshtml
CodrinCotarlanSoftwire: I thought about that. However if an error happens mid-journey (in the rare event of that), I don't think we should link the user back where he was left off. Based on [this](https://design-system.service.gov.uk/patterns/problem-with-the-service-pages/), asking for trying again is the most sensible.

DanCorderSoftwire: This is the design on the ticket, so fine here. We can raise separately if necessary

DanCorderSoftwire (approved):

PR #45, opened 2022-06-06
Seabeta 389 add off hours to request

natdeanlewissoftwire:
Tests: tested locally
Risk: minimal as change is small and this PR only adds a new request parameter in the same format as other previously working parameters
Documentation: only one comment added for assumption made, no other necessary

### DIFF
--- a/SeaPublicWebsite/Views/EnergyEfficiency/HeatingPattern.cshtml
+++ b/SeaPublicWebsite/Views/EnergyEfficiency/HeatingPattern.cshtml
@@ -76,7 +76,7 @@
                                             m => m.HoursOfHeatingMorning,
                                             new LabelViewModel
                                             {
-                                                Text = "Total number of hours in the morning (12am until 12pm)"
+                                                Text = "Total whole number of hours in the morning (12am until 12pm)"
                                             },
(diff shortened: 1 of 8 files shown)


SeaPublicWebsite/Views/EnergyEfficiency/HeatingPattern.cshtml
```diff
@@ -76,15 +76,15 @@
                                             m => m.HoursOfHeatingMorning,
                                             new LabelViewModel
                                             {
-                                                Text = "Total number of hours in the morning (12am until 12pm)"
+                                                Text = "Total whole number of hours in the morning (12am until 12pm)"
```
natdeanlewissoftwire, line 79: are these clear enough to the user that the input must be an int? the error message at the moment just says that the value is invalid if a non-integer decimal is submitted but doesn't say why.

Also, the bre api would like the total morning and hours to sum to at least 1 - if 0 is entered for both an error is thrown upon the request being sent - I'm working on this but happy to make it a separate ticket to get the core functionality merged?

DanCorderSoftwire (approved):

PR #49, opened 2022-06-08
Seabeta 261 add handler for returning users

CodrinCotarlanSoftwire:


SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -79,15 +78,13 @@ public IActionResult NewOrReturningUser_Post(NewOrReturningUserViewModel viewMod
 
                 var userDataModel = userDataStore.LoadUserData(viewModel.Reference);
```
DanCorderSoftwire, line 79: Not your code, but this line looks a bit pointless and causes a database request.

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -79,15 +78,13 @@ public IActionResult NewOrReturningUser_Post(NewOrReturningUserViewModel viewMod
 
                 var userDataModel = userDataStore.LoadUserData(viewModel.Reference);
                 
-                // TODO: Seabeta-346 to combine this with magic link logic
-                return userDataModel.UserRecommendations.Any()
-                    ? RedirectToAction("Recommendation_Get", "EnergyEfficiency", new { reference = viewModel.Reference, id = (int) userDataModel.UserRecommendations[0].Key })
-                    : RedirectToAction("NoRecommendations_Get", "EnergyEfficiency", new { reference = viewModel.Reference });
+                return RedirectToAction(nameof(ReturningUser_Get), "EnergyEfficiency",
```
DanCorderSoftwire, line 81: I'm not keen on using `RedirectToAction` here as the user will get redirected twice.
You could use `return ReturningUser_Get(viewModel.Reference);`
Or extract the logic of `ReturningUser_Get` into a separate method and call that method here.

DanCorderSoftwire: `RedirectToAction` returns a 302 and causes a new browser request so we shouldn't use it when it's not necessary

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
CodrinCotarlanSoftwire: I see what you mean by the double redirect, makes sense to just call the function.
However the ReturningUser_Get redirect is necessary as it constitutes the 'magic link' entry point when an user is emailed their reference number.

DanCorderSoftwire (approved):

PR #56, opened 2022-06-08
Seabeta 277 Update text

CodrinCotarlanSoftwire:


SeaPublicWebsite/Models/EnergyEfficiency/QuestionOptions/SolidWallsInsulated.cs
```diff
@@ -6,11 +6,11 @@ public enum SolidWallsInsulated
...
-        No,
-        [GovUkRadioCheckboxLabelText(Text = "Yes, but only some of my solid walls are insulated")]
-        Some,
-        [GovUkRadioCheckboxLabelText(Text = "Yes, all my solid walls are insulated")]
+        [GovUkRadioCheckboxLabelText(Text = "Yes, they are all insulated")]
         All,
+        [GovUkRadioCheckboxLabelText(Text = "Some are insulated and some are not")]
+        Some,
+        [GovUkRadioCheckboxLabelText(Text = "No, they are not insulated")]
+        No
```
DanCorderSoftwire, line 14: Be aware that re-ordering an enum will renumber the values. That means that (at least when we switch to the DB) saved data will be broken by this change.
I'm not sure what the save files in S3 look like, but probably safest to put this back into the same order it was in.

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Models/EnergyEfficiency/QuestionOptions/SolidWallsInsulated.cs
DanCorderSoftwire: If the design wants them in that order then we should do it. Given that we're still in dev mode it's sufficient to email the team and tell them that any saved details they have might now be broken.
If we were live you'd need to also create a database migration to update the stored values - actually that would be quite tricky to make into a non-breaking migration too.

PR #57, opened 2022-06-08
feat: Connect to database on GOV.PaaS

DanCorderSoftwire:
Set up the database connection when running on GOV.PaaS


natdeanlewissoftwire (approved):
Looks good to me!

PR #62, opened 2022-06-10
Seabeta 193 replace s3 with db

natdeanlewissoftwire:


SeaPublicWebsite/Helpers/PropertyDataHelper.cs
```diff
@@ -0,0 +1,200 @@
+﻿using SeaPublicWebsite.Data;
```
natdeanlewissoftwire, line 1: Think this is the file we were discussing this morning (previously UserDataHelper) - should we remove its use elsewhere in the code (as a separate ticket of course)?

SeaPublicWebsite.Data/DataAccessProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        context.SaveChanges();
+    }
+
+    public void UpdatePropertyData(PropertyData propertyData)
+    {
+        context.PropertyData.Update(propertyData);
+        context.SaveChanges();
+    }
+
+    public async Task<PropertyData> GetSinglePropertyData(string reference)
```
DanCorderSoftwire, line 26: Async methods should really have `Async` at the end of their names so that people remember to `await` them

SeaPublicWebsite.Data/DataAccessProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        return await context.PropertyData
+            .Include(p => p.Epc)
+            .Include(p => p.PropertyRecommendations)
+            .FirstOrDefaultAsync(p => p.Reference == reference);
```
DanCorderSoftwire, line 31: This should be `SingleAsync()`? The calling code checks that the reference is valid, so returning null here seems like a problem

SeaPublicWebsite.Data/DataStores/PropertyDataStore.cs
```diff
@@ -0,0 +1,54 @@
...
+                Reference = reference
+            };
+        }
+        return await dataAccessProvider.GetSinglePropertyData(reference.ToUpper());
+    }
+
+    public async Task<bool> IsReferenceValid(string reference)
+    {
+        var allPropertyData = await dataAccessProvider.GetAllPropertyData();
+        return allPropertyData.Any(p => p.Reference == reference.ToUpper());
```
DanCorderSoftwire, line 29: We probably need an index on the Reference column so that this doesn't get horribly slow

SeaPublicWebsite.Data/DataStores/PropertyDataStore.cs
```diff
@@ -0,0 +1,54 @@
...
+        dataAccessProvider.UpdatePropertyData(propertyData);
+    }
+
+    public async Task<string> GenerateNewReferenceAndSaveEmptyPropertyData()
+    {
+        string reference;
+        do
+        {
+            reference = RandomHelper.Generate8CharacterReference();
+        } while (await IsReferenceValid(reference));
```
DanCorderSoftwire, line 43: This is pretty inefficient, but at the moment I can't think of a way of having properly random values that doesn't do something like this...

SeaPublicWebsite.Data/DataStores/PropertyDataStore.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        PropertyData propertyData = new()
+        {
+            Reference = reference
+        };
+        
+        dataAccessProvider.AddPropertyData(propertyData);
```
DanCorderSoftwire, line 50: We also need a unique constraint on Reference as we will have multiple instances running in parallel so the check above isn't 100% guaranteed to be accurate when we reach here.

SeaPublicWebsite.Data/ErrorHandling/CustomErrorPageExceptions.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace SeaPublicWebsite.Data.ErrorHandling
```
DanCorderSoftwire, line 1: Why is this in the data project? The data project should know nothing about the website

SeaPublicWebsite.Data/ErrorHandling/ErrorHandlingFilter.cs
```diff
@@ -1,30 +1,30 @@
...
-        }
-        
-    }
-}
+﻿using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.ViewFeatures;
+
+namespace SeaPublicWebsite.Data.ErrorHandling
```
DanCorderSoftwire, line 6: This should also stay in the web project?

SeaPublicWebsite.Data/SeaPublicWebsite.Data.csproj
```diff
@@ -6,6 +6,7 @@
     </PropertyGroup>
 
     <ItemGroup>
+      <PackageReference Include="CabinetOffice.GovUkDesignSystem" Version="1.0.0-ce7e22f" />
```
DanCorderSoftwire, line 9: We really don't need this here :)

SeaPublicWebsite/Helpers/PropertyDataHelper.cs
DanCorderSoftwire: It's definitely closely related. Hopefully we can get rid of it.

DanCorderSoftwire (changes requested):

SeaPublicWebsite.Data/DataStores/PropertyDataStore.cs
DanCorderSoftwire: Also, just noticed that `GetAllPropertyData()` materialises the entire propertyData table into memory which is a bad thing (TM)

DanCorderSoftwire (approved):

DanCorderSoftwire:
This isn't ideal as we're still depending on GovUkDesignSystem in the business logic project, but we can sort that out later on

PR #65, opened 2022-06-13
Seabeta 394 Add check to ignore SAP epc reports

CodrinCotarlanSoftwire:


SeaPublicWebsite/ExternalServices/EpbEpc/EpbEpcApi.cs
```diff
@@ -88,6 +88,13 @@ public async Task<Epc> GetEpcForId(string epcId)
             }
 
             var epc = response.Data.Assessment;
+
+            // We do not surface "SAP" report data as it may interfere with the "RdSAP" ones.
```
DanCorderSoftwire, line 92: I'm not sure this comment is quite right. We ignore SAP reports because we can't convert the information in them into answers for our questions.

DanCorderSoftwire (changes requested):

SeaPublicWebsite/ExternalServices/EpbEpc/EpbEpcApi.cs
CodrinCotarlanSoftwire: I'll update the comment. Interfere doesn't quite capture what I wanted to say

PR #66, opened 2022-06-13
Seabeta 378 Update accessible loft question

CodrinCotarlanSoftwire:


CodrinCotarlanSoftwire:
On another ticket of adding a new 'Do you have access to your loft', I will make the necessary changes for the answer summary as well.

PR #67, opened 2022-06-13
Configure data protection

DanCorderSoftwire:
We need this to use form anti-forgery across multiple servers


CodrinCotarlanSoftwire (approved):

natdeanlewissoftwire (approved):
Looks fine to me - though didn't we agree not to merge migrations in anymore? Though naturally it's fine for you to as long as Codrin and I make sure we don't I'm happy with that :)

DanCorderSoftwire:
So, migrations have to be merged from their branch to main. The important thing is that two separate migration branches aren't both merged into main without one of them being fixed. The 343 branch comes from the current main head so it can be safely merged back into main.

PR #68, opened 2022-06-14
Seabeta 331 update property built year question

natdeanlewissoftwire:
Test: locally in case of epc existing/not, question being skipped/not
Risks: relies on epc data being saved to propertydata correctly, should be fine but I haven't tested this
Documentation: assumption states based on wall type/epc highlighted with a comment


SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -300,24 +300,61 @@ private static BrePropertyType GetBrePropertyType(PropertyType propertyType)
...
+                            HomeAge.From1900To1929 => "B",
+                            HomeAge.From1930To1949 => "C",
+                            HomeAge.From1950To1966 => "D",
+                            HomeAge.From1967To1975 => "E",
+                            HomeAge.From1976To1982 => "F",
+                            HomeAge.From1983To1990 => "G",
+                            HomeAge.From1991To1995 => "H",
+                            HomeAge.From1996To2002 => "I",
+                            HomeAge.From2003To2006 => "J",
+                            HomeAge.From2007ToPresent => "L",
```
natdeanlewissoftwire, line 346: no K, there is a 2007-2012 band that isn't included in the ticket, should we add this?

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -300,24 +300,61 @@ private static BrePropertyType GetBrePropertyType(PropertyType propertyType)
             }
         }
 
-        private static string GetBreConstructionDate(int? yearBuilt)
+        private static string GetBreConstructionDate(YearBuilt? yearBuilt, WallConstruction? wallConstruction, CavityWallsInsulated? cavityWallsInsulated,  HomeAge? epcConstructionAgeBand)
```
DanCorderSoftwire, line 303: As discussed the logic here will probably need to be changed a bit

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Services/RecommendationService.cs
natdeanlewissoftwire: changed to prioritise epc data over wall insulation assumptions if it exists, there are some designs incoming to potentially just use epc data if it exists for construction data and property type (and show a confirmation of this extracted data to the user rather than these question) but I think we should merge this PR first and handle that separately

SeaPublicWebsite/Services/RecommendationService.cs
```diff
@@ -311,50 +311,36 @@ private static string GetBreConstructionDate(YearBuilt? yearBuilt, WallConstruct
...
-                            HomeAge.From1983To1990 => "G",
-                            HomeAge.From1991To1995 => "H",
-                            HomeAge.From1996To2002 => "I",
-                            HomeAge.From2003To2006 => "J",
-                            HomeAge.From2007ToPresent => "L",
-                            _ => "D",
+                            CavityWallsInsulated.DoNotKnow => "D",
+                            CavityWallsInsulated.No => "D",
+                            CavityWallsInsulated.Some => "D",
+                            CavityWallsInsulated.All => "K",
```
DanCorderSoftwire, line 336: Given that we're using more accurate date ranges from the EPC should this be `I` instead of `K`?

DanCorderSoftwire (approved):

PR #72, opened 2022-06-14
Seabeta 450 Update error controller

CodrinCotarlanSoftwire:


SeaPublicWebsite/Controllers/ErrorController.cs
```diff
@@ -6,13 +6,19 @@ namespace SeaPublicWebsite.Controllers;
...
+    [HttpGet("")]
+    [HttpPost("")]
+    public IActionResult HandleException()
+    {
+        return View("ServiceIssue");
+    }
+    
     [HttpGet("{code:int}")]
-    public IActionResult Index(int code)
+    public IActionResult HandlePageNotFound(int code)
```
DanCorderSoftwire, line 17: If this is always page not found then we don't need the `code` parameter?

CodrinCotarlanSoftwire: There can be other status codes other than 404 that we do not want handled as a page not found (what codes, I do not know). That being said, the naming can be renamed to something different, any suggestions?

DanCorderSoftwire (approved):

PR #76, opened 2022-06-14
fix: Allow multiple property details to use the same EPC

DanCorderSoftwire:
The issue was that EF Core had taken the `EpcId` column as the primary key and so disallowed multiple rows for the same EPC

### DIFF
--- a/SeaPublicWebsite.Data/Migrations/20220614163252_AddDedicatedPrimaryKeyToEpcTable.cs
+++ b/SeaPublicWebsite.Data/Migrations/20220614163252_AddDedicatedPrimaryKeyToEpcTable.cs
@@ -1,0 +1,130 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+using Npgsql.EntityFrameworkCore.PostgreSQL.Metadata;
+
+#nullable disable
+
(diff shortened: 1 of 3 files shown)


SeaPublicWebsite.Data/Migrations/20220614163252_AddDedicatedPrimaryKeyToEpcTable.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<int>(
+                    name: "Id",
+                    table: "Epc",
+                    type: "integer",
+                    nullable: false,
+                    defaultValue: 0)
+                .Annotation("Npgsql:ValueGenerationStrategy", NpgsqlValueGenerationStrategy.IdentityByDefaultColumn);
```
CodrinCotarlanSoftwire, line 18: I know this is the way to go, but I have to ask: Why is it we can't update an epc with the same ID and instead have to create a new one? As in, why did the database complain?

CodrinCotarlanSoftwire (approved):

natdeanlewissoftwire (approved):

PR #90, opened 2022-06-20
Seabeta 474 Add accessibility statement page

CodrinCotarlanSoftwire:


SeaPublicWebsite/Controllers/StaticPagesController.cs
```diff
@@ -0,0 +1,12 @@
...
+namespace SeaPublicWebsite.Controllers;
+
+public class StaticPagesController : Controller
+{
+    [HttpGet("/accessibility-statement")]
+    public IActionResult AccessibilityStatement()
+    {
+        return View("AccessibilityStatement");
+    }
+}
```
CodrinCotarlanSoftwire, line 12: We had no controller to handle this. This can be used for the future privacy policy page too.

PR #94, opened 2022-06-21
Seabeta 487 capture google analytics events

ghost:


SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1126,8 +1136,27 @@ public async Task<IActionResult> AnswerSummary_PostAsync(string reference)
...
 
         [HttpGet("returning-user/{reference}")]
-        public async Task<IActionResult> ReturningUser_Get(string reference)
+        public async Task<IActionResult> ReturningUser_Get(string reference, bool fromMagicLink = true)
         {
+            if (cookieService.HasAcceptedGoogleAnalytics(Request))
+            {
+                await googleAnalyticsService.SendEvent(new GaRequestBody
+                {
+                    ClientId = Guid.NewGuid().ToString(),
```
CodrinCotarlanSoftwire, line 1145: I might be able to dig up the client ID if I look into the cookies, but it's not needed

SeaPublicWebsite/ExternalServices/GoogleAnalytics/GoogleAnalyticsService.cs
```diff
@@ -0,0 +1,46 @@
...
+    public List<GaEvent> GaEvents { get; set; }
+}
+
+public class GaEvent
+{
+    [JsonProperty(PropertyName = "name")]
+    public string Name { get; set; }
+    
+    [JsonProperty(PropertyName = "params")]
+    public object Parameters { get; set; }
```
CodrinCotarlanSoftwire, line 45: Documentation says it's an object. I could make it a `Dictionary<string, object>` and achieve the same result

SeaPublicWebsite/Views/Shared/_Layout.cshtml
```diff
@@ -1,13 +1,14 @@
...
+@using SeaPublicWebsite.ExternalServices.GoogleAnalytics
 @using SeaPublicWebsite.Helpers
 @using SeaPublicWebsite.Services
 @using SeaPublicWebsite.Services.Cookies
 @inject StaticAssetsVersioningService StaticAssetsVersioning
 @inject CookieService CookieService
+@inject GoogleAnalyticsService GoogleAnalytics
 @{
     string assetsPath = "/assets/images";
-    var bannerState = CookieService.GetAndUpdateBannerState(Context.Request, Context.Response);
```
CodrinCotarlanSoftwire, line 10: Forgot to remove this at some point

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
DanCorderSoftwire: What is the client ID used for?

Infrastructure/SetEnvironmentSecretsTemplate.sh
```diff
@@ -27,4 +27,7 @@ cf set-env "sea-beta-${PAAS_ENV_SHORTNAME}" Bre__Password "<REAL_VALUE_HERE>"
 # GovUK Notify Credentials
 cf set-env "sea-beta-${PAAS_ENV_SHORTNAME}" GovUkNotify__ApiKey "<REAL_VALUE_HERE>"
 
+# Google Analytics Credidentials
+cf set-env "sea-beta-${PAAS_ENV_SHORTNAME}" GoogleAnalytics__ApiSecret "<REAL_VALUE_HERE>"
+
```
DanCorderSoftwire, line 32: We need to run this on the environments before this ticket is merged

SeaPublicWebsite/ExternalServices/GoogleAnalytics/GoogleAnalyticsService.cs
DanCorderSoftwire: `Dictionary<string, object>` seems nicer if it works

SeaPublicWebsite/wwwroot/js/bundle/google-analytics.js
```diff
@@ -0,0 +1,139 @@
...
+            let type = input.type;
+            
+            if (type === "checkbox") {
+                if (!input.checked) {
+                    gtag("event", `change_input_${type}`, {
+                        "name": name,
+                        "value": value,
+                    });
+                }
+            } else if (type === "radio" || type === "text") {
```
DanCorderSoftwire, line 34: For text inputs does this event fire only when the control loses focus, or every time the user types a character?

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
CodrinCotarlanSoftwire: So that we know which user caused the event. I will dig up the id from the _ga cookie and use it here instead

SeaPublicWebsite/wwwroot/js/bundle/google-analytics.js
CodrinCotarlanSoftwire: Yes, that is why I used onChange. The event is only fired when focus is moved away.

Infrastructure/SetEnvironmentSecretsTemplate.sh
CodrinCotarlanSoftwire: I'll let you merge the ticket once the environment is set up

DanCorderSoftwire (approved):

PR #96, opened 2022-06-21
Seabeta 535 subtask 519.1

natdeanlewissoftwire:
subtask 1 of 2 for seabeta 519, contains accessibility changes that do not require changes to the design library


SeaPublicWebsite/Views/EnergyEfficiency/NewOrReturningUser.cshtml
```diff
@@ -65,16 +65,16 @@
 
     </div>
     <div class="govuk-grid-column-one-third">
-        <aside class="app-related-items" role="complementary">
-            <h2 class="govuk-heading-m" id="subsection-title">What does this mean?</h2>
+        <aside class="app-related-items" role="complementary" aria-labelled-by="helper text">
```
DanCorderSoftwire, line 68: I suspect this is meant to be `aria-label` instead of `aria-labelled-by`. The latter expects an element ID

SeaPublicWebsite/Views/Shared/_Layout.cshtml
```diff
@@ -101,11 +101,13 @@
                     </text>
         }))
 
-        @(RenderSection("BeforeMain", false))
 
-        <main class="govuk-main-wrapper" id="main-content" role="main">
```
DanCorderSoftwire, line 106: I realise that this is what the ticket says, however it's going directly against the GOV UK Design System guidance: https://design-system.service.gov.uk/components/back-link/

`Always place back links at the top of a page, before the <main> element. Placing them here means that the ‘Skip to main content’ link allows the user to skip all navigation links, including the back link.`

natdeanlewissoftwire: reverted

PR #97, opened 2022-06-21
Seabeta 277 revert answer changes

CodrinCotarlanSoftwire:


SeaPublicWebsite.BusinessLogic/Models/PropertyData.cs
```diff
@@ -44,6 +44,56 @@ public class PropertyData
...
+    public void ResetBackup()
+    {
+        Backup = null;
+    }
+
+    private void CopyAnswersTo(PropertyData other)
+    {
+        other.OwnershipStatus = OwnershipStatus;
+        other.Country = Country;
+        other.Epc = Epc;
```
CodrinCotarlanSoftwire, line 71: Let me know if you have a nicer way to do this (I tried iterating through the object's properties but it feels hacky)

SeaPublicWebsite.Data/Migrations/20220620125327_AddBackupObjectToPropertyData.cs
```diff
@@ -0,0 +1,36 @@
...
+                table: "PropertyData",
+                type: "integer",
+                nullable: true);
+            
+            migrationBuilder.AddForeignKey(
+                name: "FK_PropertyData_PropertyData_BackupPropertyDataId",
+                table: "PropertyData",
+                column: "BackupPropertyDataId",
+                principalTable: "PropertyData",
+                principalColumn: "PropertyDataId");
```
CodrinCotarlanSoftwire, line 22: I originally tried an 'Updates' object. The reason if failed is because it would seriously require rethinking the logic in the questionFlowService (because it would need to take into account the update object first and then the normal one).

This approach was also more flexible as I only needed to update the post methods and the AnserSummary_Get method.

SeaPublicWebsite.UnitTests/QuestionFlowServiceTests.cs
```diff
@@ -78,18 +78,6 @@ public void RunSkipLinkTestCases(QuestionFlowServiceTestCase testCase)
...
-            new Input(
-                QuestionFlowPage.Country,
-                "ABCDEFGH",
-                entryPoint: QuestionFlowPage.Country
-            ),
-            new PathByActionArguments(
-                nameof(EnergyEfficiencyController.AnswerSummary_Get),
-                "EnergyEfficiency",
-                new { reference = "ABCDEFGH" }
-            )),
```
CodrinCotarlanSoftwire, line 92: Removed change tests and actions for propery / house / bungalow / flat type, country and ownership status.

(You can still change your property type from answer summary, but you'd need to go through the whole flow. We'll fix this in one of the newer tickets detailing the new flow)

SeaPublicWebsite/Views/EnergyEfficiency/HomeAge.cshtml
```diff
@@ -50,7 +50,7 @@
                             Your @Model.Epc?.LodgementDate?.Year Energy Performance Certificate suggests your property was built
                             <strong>
                                 @(Model.Epc?.ConstructionAgeBand switch {
-                                    HomeAge.Pre1900 or HomeAge.From1900To1929 => "before 1900.",
+                                    HomeAge.Pre1900 or HomeAge.From1900To1929 => "before 1930.",
```
CodrinCotarlanSoftwire, line 53: Consistancy change

SeaPublicWebsite.BusinessLogic/Models/PropertyData.cs
```diff
@@ -44,6 +44,56 @@ public class PropertyData
     public int? HoursOfHeatingMorning { get; set; }
     public int? HoursOfHeatingEvening { get; set; }
     public decimal? Temperature { get; set; }
+    public PropertyData Backup { get; set; }
 
     public List<PropertyRecommendation> PropertyRecommendations { get; set; }
+
+    public void CreateBackup()
```
DanCorderSoftwire, line 51: `Backup` is quite loaded term so I'd like to avoid it here. I think the best we've come up with so far is `UneditedData`

DanCorderSoftwire (changes requested):

DanCorderSoftwire (approved):

PR #106, opened 2022-06-24
fix: Fix up email links

DanCorderSoftwire:
Make email links point to the right place

### DIFF
--- a/SeaPublicWebsite/Constants.cs
+++ b/SeaPublicWebsite/Constants.cs
@@ -2,5 +2,6 @@
 
 public static class Constants
 {
+    public const string FEEDBACK_URL = "https://sea-beta-dev.cloudapps.digital/feedback";
     public const string SERVICE_NAME = "Improve your property's energy efficiency";
 }
--- a/SeaPublicWebsite/appsettings.Production.json
+++ b/SeaPublicWebsite/appsettings.Production.json
@@ -10,6 +10,6 @@
     },
 
     "GovUkNotify": {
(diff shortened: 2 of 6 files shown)


SeaPublicWebsite/Constants.cs
```diff
@@ -2,5 +2,6 @@
 
 public static class Constants
 {
+    public const string FEEDBACK_URL = "https://sea-beta-dev.cloudapps.digital/feedback";
```
DanCorderSoftwire, line 5: This doesn't work on prod, but we're about to change feedback to be a separate site so then one link will make sense.

SeaPublicWebsite/appsettings.Production.json
```diff
@@ -7,5 +7,9 @@
                 "Args": { "uri": "tls://34ae15fb-5760-47f5-a1b8-93568b5df4d3-ls.logit.io:17202" }
             }
         ]
+    },
+
+    "GovUkNotify": {
+        "BaseUrl": "https://sea-beta-production.cloudapps.digital/"
```
CodrinCotarlanSoftwire, line 13: This doesn't have /energy-efficiency which seems correct, but the one below has

CodrinCotarlanSoftwire (changes requested):
Fix discrepancy between prod and dev config base url

SeaPublicWebsite/appsettings.Production.json
DanCorderSoftwire: Thanks, that's what I get for rushing

CodrinCotarlanSoftwire (approved):

PR #121, opened 2022-06-28
Seabeta 566 change your answers warning

ghost:


SeaPublicWebsite.BusinessLogic/Models/PropertyData.cs
```diff
@@ -45,6 +45,8 @@ public class PropertyData
     public int? HoursOfHeatingEvening { get; set; }
     public decimal? Temperature { get; set; }
     public PropertyData UneditedData { get; set; }
+    public bool? HasSeenRecommendations { get; set; }
+    public bool? HasEditedData { get; set; }
```
CodrinCotarlanSoftwire, line 49: HasSeenRecommendations is set to true when the user clicks 'See Results' on Answer Summary. Maybe I can rename this to 'HasConfirmedAsnwers' or 'RecommendationsRequested'?

HasEditedData is set to true from the beginning, and when the user comes in the answer summary page with property data different than the Unedited Data (i.e. a change occured).
It is set to false after requesting the recommendations.

DanCorderSoftwire: These don't need to be nullable?
`HasSeenRecommendations` seems straightforward.
For `HasEditedData` could we start it as `true`, set it to `false` when we request recommendations from BRE and only set it to `true` when we call in `DeleteUneditedData` in `UpdatePropertyAndRedirect` (if something has actually changed).
That seems simpler and I think covers all the cases?

DanCorderSoftwire (changes requested):

SeaPublicWebsite.UnitTests/PropertyDataTests.cs
```diff
@@ -64,13 +66,54 @@ public void CopiesAllAnswers()
...
+        // Assert
+        propertyData.PropertyRecommendations.Should().BeNullOrEmpty();
+    }
+    
+    public void CommitEditsKeepsPropertyRecommendationsIfThereAreNoChange()
+    {
+        // Arrange
+        var propertyData = InitializePropertyData();
+        propertyData.CreateUneditedData();
+        propertyData.WallConstruction = null;
```
DanCorderSoftwire, line 111: This is a change, so the test should fail?

SeaPublicWebsite.BusinessLogic/Models/PropertyData.cs
```diff
@@ -56,26 +55,31 @@ public void CreateUneditedData()
...
+    private bool EditedDataIsDifferent()
     {
-        if (UneditedData is null)
-        {
-            HasEditedData = true;
-            return;
-        }
-
-        HasEditedData = OwnershipStatus != UneditedData.OwnershipStatus
+        return OwnershipStatus != UneditedData.OwnershipStatus
```
DanCorderSoftwire, line 82: How will we notice if someone forgets to update this method?

DanCorderSoftwire (changes requested):

PR #131, opened 2022-07-01
Seabeta 580 epc relationship

DanCorderSoftwire:


SeaPublicWebsite.Data/Migrations/20220620125327_AddBackupObjectToPropertyData.cs
```diff
@@ -24,13 +24,13 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
-                name: "BackupPropertyDataId",
-                table: "PropertyData");
-
             migrationBuilder.DropForeignKey(
                 name: "FK_PropertyData_PropertyData_BackupPropertyDataId",
                 table: "PropertyData");
+            
+            migrationBuilder.DropColumn(
+                name: "BackupPropertyDataId",
+                table: "PropertyData");
```
CodrinCotarlanSoftwire, line 33: Does this count as changing a past migration? As in, is this reorder safe?

DanCorderSoftwire: Good question. In this case it should be safe for two reasons:
- This is in the down method and I don't think we've ever rolledback migrations on any of the environments, so this will never have been run before
- I think if you did try to run the original version it would break anyway

PR #133, opened 2022-07-01
seabeta 587 Remove primary keys from models

CodrinCotarlanSoftwire:


SeaPublicWebsite.Data/SeaDbContext.cs
```diff
@@ -41,4 +43,31 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             .HasForeignKey<PropertyData>("EditedDataId")
             .OnDelete(DeleteBehavior.Cascade);
     }
+
+    private void SetupPrimaryKeys(ModelBuilder modelBuilder)
```
DanCorderSoftwire, line 47: I think it will probably be more maintainable to have functions for each table, rather than for each thing that applies across all tables.
You could also move this into one of them:
```
modelBuilder.Entity<PropertyData>()
            .HasIndex(p => p.Reference)
            .IsUnique();
```

DanCorderSoftwire (changes requested):
Have you check that a migration isn't needed? I don't think it should be, but it would be worth checking by creating a migration and checking that it is empty

DanCorderSoftwire (approved):

PR #134, opened 2022-07-01
seabeta 495 Refactor epc parsing

CodrinCotarlanSoftwire:


SeaPublicWebsite/ExternalServices/Models/Epb/EpbEpcDto.cs
```diff
@@ -58,4 +62,537 @@ public class EpbEpcAssessmentDto
...
+
+        return null;
+    }
+    
+    private HouseType? ParseHouseType()
+    {
+        if (ParsePropertyType() is not BusinessLogic.Models.Enums.PropertyType.House)
+        {
+            return null;
+        }
```
CodrinCotarlanSoftwire, line 219: For PropertyType and HasHotWaterCylinder, property names are the same as the type name and there's a conflicting namespace, so I have to use this entire type path. I had a thought of making these methods static and give them as argument the string / Dto object, which method do you think is bette?

SeaPublicWebsite/ExternalServices/Models/Epb/EpbEpcDto.cs
```diff
@@ -58,4 +62,537 @@ public class EpbEpcAssessmentDto
...
+        
+        if (PropertyType.Contains("detached", StringComparison.OrdinalIgnoreCase))
```
DanCorderSoftwire, line 221: This will catch "semi-detached" won't it?

SeaPublicWebsite/ExternalServices/Models/Epb/EpbEpcDto.cs
DanCorderSoftwire: I think this is fine

SeaPublicWebsite/ExternalServices/Models/Epb/EpbEpcDto.cs
DanCorderSoftwire: Same for bungalows

DanCorderSoftwire (changes requested):

SeaPublicWebsite/ExternalServices/Models/Epb/EpbEpcDto.cs
CodrinCotarlanSoftwire: Ah, good catch!

PR #135, opened 2022-07-04
Seabeta 359 no epc blue banners

ghost:


SeaPublicWebsite/Views/EnergyEfficiency/GlazingType.cshtml
```diff
@@ -39,27 +39,30 @@
...
+                                    _ => throw new ArgumentOutOfRangeException()})
+                            </strong>
+                        </p>;
+                } else if (Model.YearBuilt is not null and not YearBuilt.DoNotKnow)
+                {
+                    blueBannerHint =
+                        @<p class="govuk-body">
+                            Windows in homes of your type
+                            <strong>
+                                @(Model.YearBuilt < YearBuilt.From1983To1995
```
DanCorderSoftwire, line 62: This sort of logic really doesn't belong in a cshtml file. It's business logic, so it should really be in the PropertyData class (where it can have unit tests too).
e.g. something like `bool PropertyData.HintHasSingleGlazing` which can then go into the view model and be checked here.

SeaPublicWebsite/Views/EnergyEfficiency/OutdoorSpace.cshtml
```diff
@@ -34,6 +35,17 @@
...
                         </p>
                      </text>;
+                Func<object, object> blueBannerHint = 
+                    @<p class="govuk-body">
+                        Homes of your type typically
+                        <strong>
+                            @(Model.PropertyType switch {
+                                PropertyType.House or PropertyType.Bungalow => "have some outdoor space",
+                                PropertyType.ApartmentFlatOrMaisonette => "do not have any outdoor space"
+                                })
```
DanCorderSoftwire, line 45: Maybe throw an exception for the default case

DanCorderSoftwire (changes requested):

SeaPublicWebsite/Views/EnergyEfficiency/GlazingType.cshtml
DanCorderSoftwire: Yes, sorry if that wasn't clear. This comment applies to all the similar logic in the cshtml files

PR #136, opened 2022-07-05
Seabeta 583 feedback form links

CodrinCotarlanSoftwire:


DanCorderSoftwire (approved):
As it looks like we'll be removing these feedback forms after private beta this is fine. Otherwise it would be worth trying to find a way to commonise the logic.

SeaPublicWebsite.Data/Migrations/SeaDbContextModelSnapshot.cs
```diff
@@ -281,7 +284,8 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                 {
                     b.HasOne("SeaPublicWebsite.BusinessLogic.Models.PropertyData", null)
                         .WithOne("UneditedData")
-                        .HasForeignKey("SeaPublicWebsite.BusinessLogic.Models.PropertyData", "EditedDataId");
+                        .HasForeignKey("SeaPublicWebsite.BusinessLogic.Models.PropertyData", "EditedDataId")
+                        .OnDelete(DeleteBehavior.Cascade);
```
DanCorderSoftwire, line 288: Why is this change in this commit? It should already be part of the snapshot file?

SeaPublicWebsite.Data/Migrations/SeaDbContextModelSnapshot.cs
```diff
@@ -38,7 +38,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.HasKey("Id");
 
-                    b.ToTable("DataProtectionKeys");
+                    b.ToTable("DataProtectionKeys", (string)null);
```
DanCorderSoftwire: Why is there no change here removing the `ReturningUser` column? It should have been part of the snapshot file before removing the migration

CodrinCotarlanSoftwire: I took the original file from the main branch and then created the necessary migration. Maybe that is the reason?

SeaPublicWebsite.Data/Migrations/SeaDbContextModelSnapshot.cs
CodrinCotarlanSoftwire: I'm not sure... unless it wasn't before?

PR #138, opened 2022-07-05
Seabeta 597 Update number of hours for heating bug

CodrinCotarlanSoftwire:


SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1273,7 +1273,12 @@ public async Task<IActionResult> HeatingPattern_Post(HeatingPatternViewModel vie
             }
             
             return await UpdatePropertyAndRedirect(
-                p => p.HeatingPattern = viewModel.HeatingPattern,
+                p =>
+                {
+                    p.HeatingPattern = viewModel.HeatingPattern;
+                    p.HoursOfHeatingMorning = viewModel.HoursOfHeatingMorning;
```
DanCorderSoftwire, line 1279: What if the user chooses `Other`, enters some numbers, and then chooses `All day and all night`? I think we'll save the heating hours in that case when we shouldn't

CodrinCotarlanSoftwire: In `UpdatePropertyAndRedirect`, right after `update` is called (i.e. these assignments), `ResetAllUnusedFields` is called which would delete the fields we do not want anyway.

DanCorderSoftwire (approved):

PR #158, opened 2022-07-08
feat: Change EPC HeatingType enum for better parsing and hints

CodrinCotarlanSoftwire:


SeaPublicWebsite.BusinessLogic/Models/Enums/EpcHeatingType.cs
```diff
@@ -0,0 +1,15 @@
...
+    OilBoiler,
+    LpgBoiler,
+    Storage,
+    DirectActionElectric,
+    HeatPump,
+    Other,
+    DoNotKnow,
+    Biomass,
+    CoalOrSolidFuel,
+}
```
CodrinCotarlanSoftwire, line 15: Instead of storing two heating and other heating type enums, it is much **much** easier to work with just one

SeaPublicWebsite.Data/Migrations/20220708114101_UpdateHeatingTypeColumnInEpc.cs
```diff
@@ -0,0 +1,70 @@
...
+            migrationBuilder.Sql(@"
+                UPDATE public.""Epc""
+                SET ""EpcHeatingType"" = ""Epc"".""OtherHeatingType"" + 8
+                WHERE ""Epc"".""OtherHeatingType"" IS NOT NULL AND ""Epc"".""OtherHeatingType"" != 2");
+            
+            // Then copy over the regular heating type values
+            migrationBuilder.Sql(@"
+                UPDATE public.""Epc""
+                SET ""EpcHeatingType"" = ""Epc"".""HeatingType""
+                WHERE ""Epc"".""HeatingType"" IS NOT NULL");
```
CodrinCotarlanSoftwire, line 30: Let me know if you want to have a chat about this one, it's not that straight forward (as is with any migration including merging two enum columns...)

SeaPublicWebsite.Data/Migrations/SeaDbContextModelSnapshot.cs
```diff
@@ -58,6 +58,9 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<int?>("ConstructionAgeBand")
                         .HasColumnType("integer");
 
+                    b.Property<int?>("EpcHeatingType")
+                        .HasColumnType("integer");
+
```
CodrinCotarlanSoftwire, line 63: Snapshot looks good

SeaPublicWebsite/Views/EnergyEfficiency/HeatingType.cshtml
```diff
@@ -47,19 +47,20 @@
...
-                                    HeatingType.DirectActionElectric => "electric",
-                                    HeatingType.HeatPump => "a heat pump",
+                                @(Model.Epc?.EpcHeatingType switch {
+                                    EpcHeatingType.GasBoiler => "a gas boiler.",
+                                    EpcHeatingType.OilBoiler => "an oil boiler.",
+                                    EpcHeatingType.LpgBoiler => "an LPG boiler.",
+                                    EpcHeatingType.Storage => "storage heating.",
+                                    EpcHeatingType.DirectActionElectric => "direct action electric.",
+                                    EpcHeatingType.HeatPump => "a heat pump.",
+                                    EpcHeatingType.CoalOrSolidFuel or EpcHeatingType.Biomass or EpcHeatingType.Other => "of another type.",
```
CodrinCotarlanSoftwire, line 63: Food for thought: We can expand the 'other' into 'biodiesel / community / ...' hints if we really wanted to, now that the epc enum is decoupled from the answers enum.

SeaPublicWebsite/Views/EnergyEfficiency/OtherHeatingType.cshtml
```diff
@@ -26,15 +26,16 @@
             
             @{
                 Func<object, object> blueBannerHint = null;
-                if (Model.Epc?.OtherHeatingType is not null)
+                if (Model.Epc?.EpcHeatingType is EpcHeatingType.Biomass or EpcHeatingType.CoalOrSolidFuel or EpcHeatingType.Other)
```
CodrinCotarlanSoftwire, line 29: Assuming we surface the epc recommends a user has a gas boiler, but they still proceed to 'Other heating type', then I suppose we assume they know better and not show anything here...?

SeaPublicWebsite/ExternalServices/Models/Epb/EpbEpcDto.cs
```diff
@@ -522,81 +521,203 @@ public Epc Parse()
...
+            return null;
         }
         
-        // Oil boiler check
-        // 22 - oil (community)
-        // 28 - oil (not community)
-        if (MainFuelType.Equals("22") ||
-            MainFuelType.Equals("28") ||
-            MainFuelType.Contains("oil", StringComparison.OrdinalIgnoreCase))
+        // Anything ending in '(community)'
```
DanCorderSoftwire, line 531: Is this definitely correct? There are a lot of changes below in how we parse the EPC values. If this is correct where is it documented?

SeaPublicWebsite.Data/Migrations/20220708114101_UpdateHeatingTypeColumnInEpc.cs
DanCorderSoftwire: This makes sense thanks :)

We can get away with this while the site isn't live, but it's interesting to think about how we would do this seamlessly in the future.
To be explicit, the problem here is that we are running two servers in parallel. During deployment of the first server this migration is run. Until the second server is deployed it will not be able to understand the EPC table anymore and may throw errors.

DanCorderSoftwire (changes requested):
Github seems a bit confused with the review comments

SeaPublicWebsite.Data/Migrations/20220708114101_UpdateHeatingTypeColumnInEpc.cs
```diff
@@ -0,0 +1,70 @@
...
+            migrationBuilder.Sql(@"
+                UPDATE public.""Epc""
+                SET ""HeatingType"" = ""Epc"".""EpcHeatingType""
+                WHERE ""Epc"".""EpcHeatingType"" IS NOT NULL AND ""Epc"".""EpcHeatingType"" < 8");
+            
+            migrationBuilder.Sql(@"
+                UPDATE public.""Epc""
+                SET ""OtherHeatingType"" = ""Epc"".""EpcHeatingType"" - 8
+                WHERE ""Epc"".""EpcHeatingType"" IS NOT NULL AND ""Epc"".""EpcHeatingType"" >= 8");
+
```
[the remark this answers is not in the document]
CodrinCotarlanSoftwire: I had more thought about this because it made me wonder....
Assuming the special case where we need to revert an 'other' answer into the `HeatingType` and `OtherHeatingType` columns, does that imply that the previous answers were Other and Other? If so (which I now think it's the only way to get an other response), I should add a separate sql update in the lines of `if EpcHeatingType == 7(Other) set OtherHeatingType = 2(other)`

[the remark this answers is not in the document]
DanCorderSoftwire: I think it's `if (EpcHeatingType == 7(Other) AND OtherHeatingType IS NULL) set OtherHeatingType = 2(other)`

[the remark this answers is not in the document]
CodrinCotarlanSoftwire: Ah correct! I'll update this once I delete the migration and redo it after the merge.

DanCorderSoftwire (approved):

PR #160, opened 2022-07-11
Seabeta 613 property data tests

CodrinCotarlanSoftwire:


DanCorderSoftwire (changes requested):
These tests are good, but there aren't any tests for the editing functionality

CodrinCotarlanSoftwire:
> These tests are good, but there aren't any tests for the editing functionality

We already had tests for copying data, creating the unedited data object and commiting the changes. Is that what you were referring to?

DanCorderSoftwire:
Oops, yes that was what I was talking about

PR #163, opened 2022-07-12
feat: Make BRE API more robust to unexpected responses

DanCorderSoftwire:


SeaPublicWebsite/ExternalServices/Bre/BreApi.cs
```diff
@@ -54,47 +57,53 @@ public async Task<List<BreRecommendation>> GetRecommendationsForPropertyRequestA
...
                 );
                 
-                List<BreRecommendation> recommendations = new List<BreRecommendation>();
-
-                if (response.Measures is null)
-                {
-                    return recommendations;
-                }
+                // BRE requests and responses shouldn't contain any sensitive details so we are OK to log them as-is
+                logger.LogInformation($"Received BRE response: {JsonConvert.SerializeObject(response)}");
```
natdeanlewissoftwire, line 61: Is it possible that the combination of multiple values in a response could be identifying even if an individual one wasn't. e.g. is there a postcode (or 1st half of a postcode) area where there is only one bungalow?

natdeanlewissoftwire: Naturally this applies to other places we are logging as well

natdeanlewissoftwire (approved):
Looks good pending comments :)

SeaPublicWebsite/ExternalServices/Bre/BreApi.cs
CodrinCotarlanSoftwire: Since we don't sent the postcode for now, we won't get a postcode back (obviously), so this is fine. And since the request contains only generic building information (i.e. no email, postcode, address etc.) I believe it's safe to log it.

CodrinCotarlanSoftwire (approved):

SeaPublicWebsite/ExternalServices/Bre/BreApi.cs
natdeanlewissoftwire: Yep, that's all I mean for if/when we do send postcodes :D

PR #166, opened 2022-07-12
feat: Add row versioning to database

DanCorderSoftwire:
Tested this by hammering the answer summary button to make multiple requests to BRE.
Instead of getting duplicated recommendations the error page is shown

### DIFF
--- a/SeaPublicWebsite.Data/Migrations/20220712123625_AddRowVersions.cs
+++ b/SeaPublicWebsite.Data/Migrations/20220712123625_AddRowVersions.cs
@@ -1,0 +1,55 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace SeaPublicWebsite.Data.Migrations
+{
+    public partial class AddRowVersions : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            // Note the xmin columns are built into PostgreSQL so this migration
+            // actually does nothing.
+            migrationBuilder.AddColumn<uint>(
+                name: "xmin",
+                table: "PropertyRecommendations",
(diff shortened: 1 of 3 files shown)


SeaPublicWebsite.Data/Migrations/20220712123625_AddRowVersions.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            // Note the xmin columns are built into PostgreSQL so this migration
+            // actually does nothing.
+            migrationBuilder.AddColumn<uint>(
+                name: "xmin",
+                table: "PropertyRecommendations",
+                type: "xid",
+                rowVersion: true,
+                nullable: false,
+                defaultValue: 0u);
```
CodrinCotarlanSoftwire, line 19: Looks good and simple to me.
Two questions:
1. if xmin is already in postgres, what does this migration do? Update the snapshot with that column?
2. What does xmin do? I read it has to do with concurrency control not timestamps (ticket 596).

CodrinCotarlanSoftwire (approved):

SeaPublicWebsite.Data/Migrations/20220712123625_AddRowVersions.cs
DanCorderSoftwire: So "Timestamp" in EF Core is a misleading name, it's just a unique number that increases every time that the row is changed. SQL Server has a ROWVERSION column that will do this.
In PostgreSQL things don't work quite the same, but xmin already exists and increases whenever the row is updated (but for different reasons).
The migration runs fine but doesn't actually affect the xmin column on `Up` or `Down`

natdeanlewissoftwire (approved):

PR #170, opened 2022-07-13
Seabeta 639 refactor controller methods

DanCorderSoftwire:
This is the first step in refactoring the controller classes. Here we're splitting apart the business logic and website logic that's currently mixed together in the QuestionFlowService.

I've run through the flow, including going back while finding an EPC and changing answers from the answer summary and check your answers pages.


SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -214,28 +212,28 @@ public async Task<IActionResult> FindEpc_Post(FindEpcViewModel viewModel)
                 return await FindEpc_Get(viewModel.Reference);
             }
 
-            var propertyData = await propertyDataStore.LoadPropertyDataAsync(viewModel.Reference);
```
DanCorderSoftwire, line 217: This change wasn't strictly necessary, but it makes things more consistent

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -454,19 +454,16 @@ public async Task<IActionResult> NoEpcFound_Post(NoEpcFoundViewModel viewModel)
             {
                 return await NoEpcFound_Get(viewModel.Reference);
             }
-            
-            var propertyData = await propertyDataStore.LoadPropertyDataAsync(viewModel.Reference); 
-            
-            var forwardArgs = questionFlowService.ForwardLinkArguments(QuestionFlowPage.ConfirmEpcDetails, propertyData);
-            return RedirectToAction(forwardArgs.Action, forwardArgs.Controller, forwardArgs.Values);
+
+            return await UpdatePropertyAndRedirect(p => { }, viewModel.Reference, QuestionFlowStep.NoEpcFound);
```
DanCorderSoftwire, line 458: Note that this used to be `ConfirmEpcDetails` which I think was a mistake (although it went to the same page which I guess is why it wasn't spotted earlier)

SeaPublicWebsite.BusinessLogic/Services/QuestionFlowService.cs
```diff
@@ -0,0 +1,806 @@
...
+
+        private QuestionFlowStep YourRecommendationsBackDestination()
+        {
+            return QuestionFlowStep.AnswerSummary;
+        }
+
+        private QuestionFlowStep NewOrReturningUserForwardDestination()
+        {
+            // TODO: Routing for the next step of returning user?
+            return QuestionFlowStep.Country;
```
CodrinCotarlanSoftwire, line 446: TODO still here. Any fix for this?

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -271,13 +269,15 @@ public async Task<IActionResult> AskForPostcode_Post(AskForPostcodeViewModel vie
...
+            var nextStep = questionFlowService.NextStep(QuestionFlowStep.AskForPostcode, propertyData);
+            var forwardArgs = GetActionArgumentsForQuestion(nextStep, viewModel.Reference);
 
             return RedirectToAction(
                 forwardArgs.Action,
                 forwardArgs.Controller,
-                // TODO: seabeta-579 we'd rather use the values from the question flow service
+                // It would be nice to find a way to get GetActionArgumentsForQuestion() to include the postcode and
+                // house number.
                 new { reference = viewModel.Reference, postcode = viewModel.Postcode, number = viewModel.HouseNameOrNumber });
```
CodrinCotarlanSoftwire, line 281: First thing I could think of is passing both values as parameters to `GetActionArgumentsForQuestion()`. The logic might get complicated because anonymous object creation is quite strict regarding types, see https://stackoverflow.com/questions/27702704/include-properties-based-on-conditions-in-anonymous-types...

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
natdeanlewissoftwire: Ah yep, that's probably on me, definitely a mistake, good spot!

natdeanlewissoftwire (approved):

SeaPublicWebsite.BusinessLogic/Services/QuestionFlowService.cs
DanCorderSoftwire: Nope, not as part of this refactoring anyway.
Actually, I'll remove this comment as some of the target pages are outside the question flow anyway so it wouldn't make sense to return then from here.

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
CodrinCotarlanSoftwire: Since this does nothing, we can modify the `.cshtml` file so that the button has `href=<nextPageURL>`. That will make the govuk 'button' an `<a>` element, removing the need for an unnecessary post.

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -676,19 +673,16 @@ public async Task<IActionResult> CheckYourUnchangeableAnswers_Post(CheckYourUnch
                 return await CheckYourUnchangeableAnswers_Get(viewModel.Reference, viewModel.EntryPoint);
             }
 
-            var propertyData = await propertyDataStore.LoadPropertyDataAsync(viewModel.Reference);
-
-            var forwardArgs = questionFlowService.ForwardLinkArguments(QuestionFlowPage.CheckYourUnchangeableAnswers, propertyData, viewModel.EntryPoint);
-            return RedirectToAction(forwardArgs.Action, forwardArgs.Controller, forwardArgs.Values);
+            return await UpdatePropertyAndRedirect(p => { }, viewModel.Reference, QuestionFlowStep.CheckYourUnchangeableAnswers);
```
CodrinCotarlanSoftwire, line 676: Same Idea here

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1643,10 +1638,8 @@ public async Task<IActionResult> AlternativeRecommendations_Post(AlternativeReco
         }
 
         [HttpGet("water-and-energy-efficiency/{reference}")]
-        public async Task<IActionResult> WaterAndElectricityEfficiency_Get(string reference, bool fromNoRecommendations, bool fromActionPlan = false)
+        public IActionResult WaterAndElectricityEfficiency_Get(string reference, bool fromNoRecommendations, bool fromActionPlan = false)
         {
-            var propertyData = await propertyDataStore.LoadPropertyDataAsync(reference);
```
CodrinCotarlanSoftwire, line 1648: I mayhave forgotten this here on a previous ticket, thanks for clearing it out

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1819,5 +1814,87 @@ private void TrySendReferenceNumberEmail(string emailAddress, string reference)
...
+            return GetActionArgumentsForQuestion(backStep, propertyData?.Reference, entryPoint);
+        }
+
+        private PathByActionArguments GetActionArgumentsForQuestion(QuestionFlowStep question, string reference = null, QuestionFlowStep? entryPoint = null)
+        {
+            return question switch
+            {
+                QuestionFlowStep.Start => new PathByActionArguments(nameof(Index), "EnergyEfficiency"),
+                QuestionFlowStep.NewOrReturningUser => new PathByActionArguments(nameof(NewOrReturningUser_Get), "EnergyEfficiency"),
+                QuestionFlowStep.OwnershipStatus => new PathByActionArguments(nameof(OwnershipStatus_Get), "EnergyEfficiency", GetPathValues(reference, entryPoint)),
```
CodrinCotarlanSoftwire, line 1833: Some of these (for example Ownership Status) are unreachable with an entry point. Please double check the others.

CodrinCotarlanSoftwire (changes requested):
Small note on unnecessary post requests, may as well fix them in this ticket.
Looks good otherwise!

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1823,73 +1805,95 @@ private void TrySendReferenceNumberEmail(string emailAddress, string reference)
...
+            QuestionFlowStep? entryPoint = null)
         {
             if (reference == null)
             {
                 throw new ArgumentException("Reference must be provided");
             }
-            
-            if (entryPoint == null)
+
+            // If entryPoint is null then it won't appear in the URL
```
CodrinCotarlanSoftwire, line 1876: Wonderful!

PR #172, opened 2022-07-14
Seabeta 639 add answer service

DanCorderSoftwire:
This should all just be moving things around.
The aim is to simplify the controller logic - ideally a controller method will just call single service method.
This also splits the database and business code when doing answer updates


SeaPublicWebsite.BusinessLogic/Services/QuestionFlowService.cs
```diff
@@ -471,11 +471,7 @@ private QuestionFlowStep FindEpcForwardDestination(PropertyData propertyData)
 
         private QuestionFlowStep AskForPostcodeForwardDestination(PropertyData propertyData)
```
CodrinCotarlanSoftwire, line 472: Can remove the `propertyData` parameter then

CodrinCotarlanSoftwire (approved):

PR #185, opened 2022-07-25
feat: name antiforgery token explicitly and hide kestrel header

natdeanlewissoftwire:
tested locally
minimal risk - this change is only cosmetic as far as I can see from the relevant docs
no documentation required

### DIFF
--- a/SeaPublicWebsite/Program.cs
+++ b/SeaPublicWebsite/Program.cs
@@ -22,6 +22,7 @@ namespace SeaPublicWebsite
             
             var builder = WebApplication.CreateBuilder(args);
             
+            //Hide that we are using Kestrel for security reasons
             builder.WebHost.ConfigureKestrel(serverOptions => serverOptions.AddServerHeader = false);
             
             var startup = new Startup(builder.Configuration, builder.Environment);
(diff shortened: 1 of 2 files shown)


SeaPublicWebsite/Program.cs
```diff
@@ -21,6 +22,8 @@ public static void Main(string[] args)
             
             var builder = WebApplication.CreateBuilder(args);
             
+            builder.WebHost.ConfigureKestrel(serverOptions => serverOptions.AddServerHeader = false);
```
Binney, line 25: LGTM, worth sticking in a citation for why this is being removed? or just a comment "Hide what server we are using for security reasons"

SeaPublicWebsite/Startup.cs
```diff
@@ -112,6 +112,7 @@ private void ConfigureCookieService(IServiceCollection services)
         {
             services.Configure<CookieServiceConfiguration>(
                 configuration.GetSection(CookieServiceConfiguration.ConfigSection));
+            services.AddAntiforgery(options => options.Cookie.Name = "Antiforgery");
```
Binney, line 115: Similar to the above really, is it worth a comment on why this is here?

Binney (approved):
Perfecto :+1:

PR #186, opened 2022-07-25
fet: add security headers

natdeanlewissoftwire:
Tested locally
Risks: I've not yet had much time to read fully into the docs for these changes, would another pair of eyes looking through to make sure I'm using these correctly :) In general though no huge changes have been made, it's possible the content security policy header has disabled some external functionality I wasn't aware we were using.
No documentation required

### DIFF
--- a/SeaPublicWebsite/Middleware/SecurityHeadersMiddleware.cs
+++ b/SeaPublicWebsite/Middleware/SecurityHeadersMiddleware.cs
@@ -1,0 +1,23 @@
+﻿using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace SeaPublicWebsite.Middleware;
+
+public class SecurityHeadersMiddleware
+{
+    private readonly RequestDelegate next;
+    
+    public SecurityHeadersMiddleware(RequestDelegate next)
+    {
+        this.next = next;
+    }
+
+    public Task Invoke(HttpContext context)
+    {
+        context.Response.Headers.Add("X-Content-Type-Options", "nosniff");
+        context.Response.Headers.Add("Content-Security-Policy", "default-src 'self");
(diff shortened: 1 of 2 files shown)


SeaPublicWebsite/Middleware/SecurityHeadersMiddleware.cs
```diff
@@ -0,0 +1,23 @@
...
+    
+    public SecurityHeadersMiddleware(RequestDelegate next)
+    {
+        this.next = next;
+    }
+
+    public Task Invoke(HttpContext context)
+    {
+        context.Response.Headers.Add("X-Content-Type-Options", "nosniff");
+        context.Response.Headers.Add("Content-Security-Policy", "default-src 'self'; script-src * 'unsafe-inline' 'unsafe-eval'");
```
natdeanlewissoftwire, line 18: script source here is unsafe, but I need the govuk design component for a conditional text input depending on a checkbox, and also I believe google analytics needs its script to be allowed. Couldn't work out how to just allow the behaviour we want so this allows everything at the moment. Not sure how much of a security vulnerability this is.

Binney: In an ideal world we'd explicitly specify those 2 script sources as safe, based on their domains. Do we always know where the scripts will be coming from?

Having a robust CSP is great if you aren't relying on loads of external scripts so I'm definitely in favour if possible

Binney (commented):
Imo it'd be good to add a more restrictive CSP

SeaPublicWebsite/Middleware/SecurityHeadersMiddleware.cs
natdeanlewissoftwire: In discussion with BEIS this is not a priority so I've added a tech debt ticket for it, am merging now and will come back to this when I get time :)

PR #209, opened 2022-09-05
fix: Add loft insulation recommendation A1 alongside A

CodrinCotarlanSoftwire:


CodrinCotarlanSoftwire:
Recommendation A was not sent for years before 1967, but A1 instead, so we needed to support that as well.

davidm-m (approved):
This looks good. I'd want to double check with the client that the summary should be the same for both A and A1, but I think it's right.

Is there a chance that the calculator can return A and A1, and so we'll show two identical recommendations?

CodrinCotarlanSoftwire:
The non-applicable measure will be filtered out automatically by the BRE tool, so we will always get at most one of those back.
I will ask if there is any difference between the measures, but from the API we only use installation costs, lifetime and saving amounts, so I don't think so.

CodrinCotarlanSoftwire:
Spoke with Alec, we can make both A and A1 reference the current loft insulation recommendation. There's a separate copy for A1 not included in the code yet, so I'll create a separate ticket to add that.

PR #216, opened 2022-10-14
BEISSEA-28: Fix heating type parse error

CodrinCotarlanSoftwire:


SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
```diff
@@ -684,7 +684,7 @@ private static bool HasSolidWalls(string description)
             
             if (MainHeatingDescription is null)
             {
-                return EpcHeatingType.DoNotKnow;
+                return null;
```
CodrinCotarlanSoftwire, line 687: DoNotKnow is something the user selects when they don't know the answer, not a parsed value from the epc!

PR #229, opened 2023-05-26
BEISSEA-67 Generate recommendations and action plan PDFs

kirstenland:
Current PDFs:
[Recommendations.pdf](https://github.com/UKGovernmentBEIS/beis-simple-energy-advice-beta/files/11574786/Recommendations.53.pdf)
[ActionPlan.pdf](https://github.com/UKGovernmentBEIS/beis-simple-energy-advice-beta/files/11574787/ActionPlan.27.pdf)

Remaining to do / risks in different environments:
 * Add basic auth username and password secrets to production environment
 * Make sure everything exists to make puppeteer work on deployed environments. It just worked on my machine but there may be dependencies to install on linux machines. Maybe can test in docker?
 * Check whether we can access with puppeteer on localhost in deployed environments. There could be certificate issues as we have had these locally


SeaPublicWebsite/Startup.cs
```diff
@@ -1,187 +1,194 @@
-using System;
```
davidm-m, line 1: Looks like this file was changed to LF line endings somehow - can you change it back to CRLF please?

SeaPublicWebsite/Models/EnergyEfficiency/ReturnToActionPlanViewModel.cs
```diff
@@ -8,4 +8,5 @@ public class ReturnToActionPlanViewModel
     [GovUkValidateRequired(ErrorMessageIfMissing = "Enter a valid email address")]
     public string EmailAddress { get; set; }
     public bool EmailSent { get; set; }
+    public bool PdfVersion { get; set; }
```
davidm-m, line 11: Not sure I like `PdfVersion` as a variable name, I keep thinking it will be a version number rather than a boolean. Would `IsPdf` work? Or do you have any other ideas?

SeaPublicWebsite/Views/EnergyEfficiency/ActionPlan/ActionPlanWithSavedRecommendations.cshtml
```diff
@@ -141,7 +151,7 @@
                        new()
                        {
                            Classes = "sea-saved-recommendations-table-headers",
-                           Html = @<a class="govuk-link" href=@Url.Action("Recommendation_Get", new {id = (int) recommendation.Key, reference = Model.PropertyData.Reference, fromActionPlan = true})>
+                           Html = @<a class="govuk-link" href="@Model.UrlPrefix@Url.Action("Recommendation_Get", new {id = (int) recommendation.Key, reference = Model.PropertyData.Reference, fromActionPlan = true})">
```
davidm-m, line 154: At first glance having the two statements next to each other with no space looks wrong to me. It works but it's not immediately clear what it's doing. Instead of `href="@foo@bar"` could we do something like `href="@{foo + bar}"`? I think that would be more understandable.

SeaPublicWebsite/Views/EnergyEfficiency/ActionPlan/Partials/_DiscardedRecommendations.cshtml
```diff
@@ -11,7 +11,7 @@
                    @foreach(var recommendation in Model.GetDiscardedRecommendations())
                    {
                        <li>
-                           <a href="@Url.Action("Recommendation_Get", new {id = (int) recommendation.Key, reference = Model.PropertyData.Reference, fromActionPlan = true})">
+                           <a class="govuk-link" href="@Model.UrlPrefix@Url.Action("Recommendation_Get", new {id = (int) recommendation.Key, reference = Model.PropertyData.Reference, fromActionPlan = true})">
```
davidm-m, line 14: Same here about the two statements right next to each other.

SeaPublicWebsite/Views/Shared/_PdfLayout.cshtml
```diff
@@ -0,0 +1,63 @@
...
+@using SeaPublicWebsite.Services
+@inject StaticAssetsVersioningService StaticAssetsVersioning
+@inject FullHostnameService FullHostnameService
+@{
+    string assetsPath = "/assets/images";
+}
+<!DOCTYPE html>
+
+<!-- Build number: @(BuildNumberHelper.GetBuildNumber()) -->
+<!-- Server name: @(Environment.MachineName) -->
```
davidm-m, line 13: We presumably don't need these comments?

SeaPublicWebsite/Views/Shared/_Recommendation.cshtml
```diff
@@ -1,131 +1,69 @@
 ﻿@using GovUkDesignSystem
 @using GovUkDesignSystem.GovUkDesignSystemComponents
-@using SeaPublicWebsite.Controllers
-@using SeaPublicWebsite.Models.EnergyEfficiency
-@model SeaPublicWebsite.Models.EnergyEfficiency.RecommendationViewModel;
+@model SeaPublicWebsite.Models.EnergyEfficiency.SingleRecommendationDetailsViewModel;
 
-@{
-    ViewBag.Title = "Recommendation " + (Model.RecommendationIndex + 1) + " of " + Model.PropertyRecommendations.Count;
```
davidm-m, line 8: Does this page not get a title any more? Or is it being set elsewhere?

SimpleEnergyAdvice.sln
```diff
@@ -1,109 +1,109 @@
-﻿
```
davidm-m, line 1: Another file that's changed to LF line endings for some reason.

davidm-m (changes requested):
All the large-scale stuff looks great. I love the recommendations refactor! Very neat and DRY.
I have got a few smaller comments that would be great if you could look into.

SeaPublicWebsite/Views/Shared/_Recommendation.cshtml
kirstenland: There has been a bit of a reshuffle so that the _Recommendation shared layout only covers rendering the core chunk of the recommendations page that is shared across the pdf and the html page.

YourRecommendation.cshtml contains the rest of the html page, including the title and the layout.

The page that will become the pdf doesn't have a title set as I didn't think it was necessary though happy to set one if it does anything.

davidm-m: Awesome, thanks for explaining. I don't expect the pdf needs one so fine to leave as-is

kirstenland:
Pictures of the buttons

image
image
image

SeaPublicWebsite/Models/EnergyEfficiency/ActionPlanViewModel.cs
```diff
@@ -48,5 +48,10 @@ public string GetSavingText(PropertyRecommendation recommendation)
         {
             return recommendation.Key == RecommendationKey.InstallHeatPump ? "-" : $"£{recommendation.Saving:N0}";
         }
+
+        public string GetUrlWithPrefix(string url)
+        {
+            return $"{UrlPrefix}{url}";
+        }
```
davidm-m, line 55: Great idea, this is so much nicer than before.

SeaPublicWebsite/Views/EnergyEfficiency/ActionPlan/Partials/_ActionPlanBeforeMainLinks.cshtml
```diff
@@ -0,0 +1,25 @@
+﻿@using GovUkDesignSystem
+@using GovUkDesignSystem.GovUkDesignSystemComponents
+@using SeaPublicWebsite.BusinessLogic.Models.Enums
+@using SeaPublicWebsite.Models.EnergyEfficiency
+@model SeaPublicWebsite.Models.EnergyEfficiency.ActionPlanViewModel;
+
+@if (!Model.IsPdf)
```
davidm-m, line 7: I find it a bit weird that we've got a view that's sometimes empty. I think it would make more sense to keep the `IsPdf` check on the main view rather than this partial.
That means we're repeating ourselves a bit but I prefer to know that if I see this partial being used on a page then its contents will actually show up.

davidm-m (changes requested):
I'm a bit concerned about having a button that looks like a link. Would be interested in hearing your thoughts on that.

SeaPublicWebsite/Views/EnergyEfficiency/Partials/_DownloadActionPlan.cshtml
```diff
@@ -2,5 +2,8 @@
 
 @using (Html.BeginForm("GenerateActionPlanPdf_Post", "EnergyEfficiency", new { reference = Model.Reference }))
 {
-    <input type="submit" value="Download Action Plan PDF"/>
+    <button type="submit" class="govuk-link sea-download-button @Model.ButtonClass">
+        <img src="/assets/images/download-icon.png" aria-hidden="true" />
+        <span>Download Action Plan PDF</span>
+    </button>
```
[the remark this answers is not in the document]
kirstenland: It could be an actual link though that would require changing the endpoint that creates and returns the Action Plan pdf from a POST request to a GET request. I don't think that's actually a problem as it does GET a thing and doesn't do anything POST-ish, I just wasn't sure what the conventions were here.

There are other buttons pretending to be links on BEIS (e.g. "Cancel and continue without an EPC") but I'm not sure that makes it a good idea - I think I'd prefer it was link-styled rather than button-styled to avoid redesigning things but I'm happy to change it to an actual link instead of a pretend link.

davidm-m (approved):
Lovely stuff.

SeaPublicWebsite/appsettings.DEV.json
```diff
@@ -11,5 +11,9 @@
 
     "GoogleAnalytics": {
         "MeasurementId": "G-RJC885GHSW"
+    },
+
+    "SimpleEnergyAdvice": {
+        "BaseUrl": "https://sea-beta-DEV.cloudapps.digital"
```
SamSeed-Softwire, line 17: Have yet to test whether is picked up correctly - will see if it works correctly when deploying to dev and whether puppeteer works then too.

SamSeed-Softwire (commented):
Still a bit WIP, but wanted to get people's thoughts before continuing! Thanks :)

SeaPublicWebsite/appsettings.DEV.json
LadunOmideyiSoftwire: Are you adding the live url to prod as well?

SeaPublicWebsite/wwwroot/css/_simple-energy-advice-custom.scss
```diff
@@ -311,6 +311,35 @@
     font-weight: 700;
 }
 
+.sea-download-button {
```
kirstenland, line 314: We presumably don't need all this CSS anymore - we should make sure we have only the css we need

SeaPublicWebsite/appsettings.DEV.json
SamSeed-Softwire: I made a similar change in `SeaPublicWebsite/appsettings.Production.json` but with the live URL if that's what you mean?

SeaPublicWebsite/wwwroot/css/_simple-energy-advice-custom.scss
kirstenland: We should also be able to remove the download-icon.png file from wwwroot/assets/images

SeaPublicWebsite/Views/EnergyEfficiency/Partials/_DownloadActionPlan.cshtml
```diff
@@ -0,0 +1,11 @@
+﻿@using GovUkDesignSystem
+@using GovUkDesignSystem.GovUkDesignSystemComponents
+@model SeaPublicWebsite.Models.EnergyEfficiency.DownloadPdfButtonViewModel
+
+@using (Html.BeginForm("GenerateActionPlanPdf_Post", "EnergyEfficiency", new { reference = Model.Reference }))
+{
+    @await Html.GovUkButton(new ButtonViewModel
```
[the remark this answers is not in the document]
davidm-m: We definitely want to move the top-of-screen button down a bit, it looks odd stuck to the divider there. No comment on the lower down bit, it looks fine to me.

davidm-m (changes requested):
Code looks fine to me! I agree with Kirsty that we can ditch that custom CSS and the icon now, and we should move the placement of the buttons a little too.

SeaPublicWebsite/Views/EnergyEfficiency/Partials/_DownloadActionPlan.cshtml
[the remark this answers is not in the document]
davidm-m: Ah yeah, the distance between "Download Recommendations" and "Change your answers" should be the same as the distance between "Email a link to your Action Plan" and "Download your information".

SeaPublicWebsite/wwwroot/css/_simple-energy-advice-custom.scss
SamSeed-Softwire: Done!

SeaPublicWebsite/Views/EnergyEfficiency/Partials/_DownloadActionPlan.cshtml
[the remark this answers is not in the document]
SamSeed-Softwire: Should be sorted now!

davidm-m (approved):
Great stuff 👍

PR #231, opened 2023-06-16
BEISSEA-67 Use CloudFoundry's apt buildpack to install Chrome's dependencies when an instance starts

davidm-m:
This is, I think, the right way to do this. CloudFoundry's apt buildpack is apparently "experimental" but it's been around for 6 years and is still maintained. All it does is install the dependencies you list in `apt.yml`.

### DIFF
--- a/SeaPublicWebsite/apt.yml
+++ b/SeaPublicWebsite/apt.yml
@@ -1,0 +1,16 @@
+---
+packages:
+  - libnss3
+  - libnssutil3
+  - libsmime3
+  - libnspr4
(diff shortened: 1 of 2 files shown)


kirstenland (approved):
Seems worth a try

LadunOmideyiSoftwire (approved):
This is definitely the way recommended in the gov.uk paas docs!

davidm-m:
> This is definitely the way recommended in the gov.uk paas docs!

Is it?! I looked through the docs for something like this and couldn't find it 🙃

PR #237, opened 2023-07-03
Beissea 67 pdf fix branch

xonde:


SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1319,14 +1322,16 @@ public async Task<IActionResult> YourRecommendations_Post(YourRecommendationsVie
         [HttpPost("your-recommendations-download/{reference}")]
         public async Task<IActionResult> GenerateRecommendationsPdf_Post(string reference)
         {
-            var stream = await pdfGenerationService.GeneratePdf($"energy-efficiency/pdf-generation/your-recommendations/{reference}");
+            var encodedReference = HttpUtility.HtmlEncode(reference);
```
davidm-m, line 1325: I've just realised now this is the wrong thing to do - since this is going to be part of a url we should be calling the `UrlEncode` method. It's on the same class so a tiny fix.

davidm-m (changes requested):
Just one tiny change that's totally my fault, sorry!

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
xonde: Good catch 🤦🏾‍♂️

davidm-m (approved):
Awesome! Go ahead and merge it

PR #239, opened 2023-07-20
BEISSEA-76: Prevent Test Project from being published

xonde:
Revert to using Newtonsoft


SeaPublicWebsite.UnitTests/SeaPublicWebsite.UnitTests.csproj
```diff
@@ -4,6 +4,7 @@
         <TargetFramework>net6.0</TargetFramework>
         <IsPackable>false</IsPackable>
         <RootNamespace>Tests</RootNamespace>
+        <IsPublishable>false</IsPublishable>
```
xonde, line 7: This is the main change

PR #240, opened 2023-07-27
Remove HTTPS redirection

davidm-m:
Required by the new DESNZ platform. I don't expect you'll have any comments but thought it worth getting a second pair of eyes anyway!

This PR is set to merge into the new dev branch, and I'll also merge it into staging. That way we won't have HTTPS redirection on the branches we're testing the new platform with.


xonde:
A lot of code to look through here, but managed to get it done. LGTM 🚀

xonde (approved):

PR #241, opened 2023-07-27
BEISSEA-NONE: Update Dockerfile to include npm install

xonde:


davidm-m (approved):
I should really have tried building your first docker branch before approving it 😅

PR #242, opened 2023-08-31
Fix Puppeteer on new platform

davidm-m:
Copied the list of packages we needed to install on gov PaaS to the Dockerfile


xonde (approved):
LGTM!

PR #247, opened 2023-09-06
BEISSEA-NONE Move Chrome installation to base image instead of build image

davidm-m:
Three key things here:
* We now install Chrome in Docker, not in .NET
* Chrome needs to be installed in the image we actually run .NET from, not the build image (this took me the longest to work out)
* Our old method for getting the local address started returning `http://[::]:80` which Chrome doesn't recognise as a valid address, so I've hardcoded it. I'm not very happy about doing that but since everything's in Docker the address should always be the same.

### DIFF
--- a/SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
+++ b/SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
@@ -1,9 +1,5 @@
-﻿using System;
-using System.IO;
-using System.Linq;
+﻿using System.IO;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Hosting.Server;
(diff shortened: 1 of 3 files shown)


SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
```diff
@@ -58,6 +47,6 @@ public async Task<Stream> GeneratePdf(string path)
 
     private string GetLocalAddress()
     {
-        return server.Features.Get<IServerAddressesFeature>().Addresses.First();
+        return "http://localhost:80";
```
xonde, line 50: We should leave a comment here, mentioning that port 80 is the port the app runs on in the docker container, and to remember to change it here if the docker file setting are ever changed

xonde (changes requested):
Looks all good, just a small update

PR #248, opened 2023-09-14
Beissea 85 pen test fixes

xonde:


SeaPublicWebsite/Services/Cookies/CookieService.cs
```diff
@@ -90,6 +90,6 @@ public void SetCookie<T>(HttpResponse response, string cookieName, T cookie)
         response.Cookies.Append(
             cookieName,
             cookieString,
-            new CookieOptions {Secure = true, SameSite = SameSiteMode.Lax, MaxAge = TimeSpan.FromDays(Configuration.DefaultDaysUntilExpiry)});
+            new CookieOptions {Secure = true, SameSite = SameSiteMode.Lax, MaxAge = TimeSpan.FromDays(Configuration.DefaultDaysUntilExpiry), HttpOnly = true});
```
davidm-m, line 93: The report makes it look like all the cookies had `httponly` already, and it's just that the built in anti forgery token didn't have `secure`. Have you checked for that?

davidm-m (commented):
CSP stuff look great, I just want to check that the change you've made to cookies is the right one.

davidm-m (approved):
Lovely stuff 👍

davidm-m:
Just realised: **PLEASE DON'T ACTUALLY MERGE THIS!**
Merging into main triggers a production deployment. We should be merging into dev first.

xonde:
Just to note, while this should _technically_ work, The way it's set up on the new platform means we're always in a dev environment. So I think we should sort out BEISSEA-86 before we merge this. I'll ask Jeni if I can prioritise that after doing the data capture ticket

davidm-m:
> Just to note, while this should _technically_ work, The way it's set up on the new platform means we're always in a dev environment. 

Ha, I was just going to write an identical comment. Sounds good to me.

SeaPublicWebsite/Startup.cs
```diff
@@ -117,10 +120,16 @@ private void ConfigureCookieService(IServiceCollection services)
...
+            services.AddAntiforgery(ConfigureAntiForgeryCookieOptions);
             services.AddScoped<CookieService, CookieService>();
         }
 
+        private void ConfigureAntiForgeryCookieOptions(AntiforgeryOptions options)
+        {
+            options.Cookie.Name = "Antiforgery";
+            options.Cookie.SecurePolicy = webHostEnvironment.IsDevelopment() ? CookieSecurePolicy.None : CookieSecurePolicy.Always;
+        }
+
```
davidm-m, line 132: Turns out HER has this exact change and it's breaking the health check since the health check is apparently over HTTP, not HTTPS. We'll need to add in an exception to the health check endpoint that it doesn't use an anti forgery cookie.

davidm-m (changes requested):
Making the anti forgery cookie secure broke HER, so we'll need to make an additional change to avoid that. I've put more details in a comment.

SeaPublicWebsite/Startup.cs
xonde: Sounds good, thanks for checking

PR #250, opened 2023-09-26
BEISSEA-86: Remove GovUK PaaS database URL check

xonde:


SeaPublicWebsite/Startup.cs
```diff
@@ -82,36 +82,11 @@ private void ConfigureGoogleAnalyticsService(IServiceCollection services)
         private void ConfigureDatabaseContext(IServiceCollection services)
         {
             var databaseConnectionString = configuration.GetConnectionString("PostgreSQLConnection");
-            if (!webHostEnvironment.IsDevelopment())
-            {
-                // In Gov.PaaS the Database URL is automatically put into the DATABASE_URL environment variable
-                databaseConnectionString = DatabaseUrlToConnectionString(configuration["DATABASE_URL"]);
-            }
-
```
xonde, line 90: Removing this will break GovUK PaaS hosted sites, I'm not sure if we would want to add some `IsGovPaaSHosted()` function to have it work on both services until Gov.PaaS is shut down or we deactivate

davidm-m: I'm confident in saying we'll never need to switch back to gov PaaS hosting, so I don't think we need anything like that. We also won't be making any prod releases to it so in the extremely unlikely case we do want to switch hosting back instead of fixing forward on the new infra, this change won't affect PaaS.

davidm-m (approved):

davidm-m:
Don't forget this should be merged into dev before main!

PR #257, opened 2023-12-18
Pc 576 fix duplicate address

xonde:
I've added a few reviewers as I'm not sure who specifically to add, apologies!


SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -257,10 +256,12 @@ public async Task<ViewResult> ConfirmAddress_Get(string reference, string postco
         {
             List<EpcSearchResult> epcSearchResults = await epcApi.GetEpcsInformationForPostcodeAndBuildingNameOrNumber(postcode, number);
             
-            var filteredEpcSearchResults = epcSearchResults.Where(e =>
+            var matchingEpcSearchResults = epcSearchResults.Where(e =>
                 e.Address1.Contains(number, StringComparison.OrdinalIgnoreCase) ||
                 e.Address2.Contains(number, StringComparison.OrdinalIgnoreCase)).ToList();
 
+            var filteredEpcSearchResults = await FilterExpiredEpc(matchingEpcSearchResults);
```
jamiehumphries, line 263: Perhaps I'm misunderstanding, but is there a possible simpler version of this that just does the filtering on the results as they already are, without additional database checks? What's the reason for not just picking out unique addresses from the `matchingEpcSearchResults` variable at this stage?

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcApi.cs
```diff
@@ -62,8 +62,8 @@ public EpbEpcApi(IOptions<EpbEpcConfiguration> options, IMemoryCache memoryCache
 
             return EpcSearchResult.SortEpcsInformation(epcsInformation);
         }
-
-        public async Task<Epc> GetEpcForId(string epcId)
+        
+        public async Task<EpbEpcAssessmentDto> GetEpcDtoForId(string epcId)
```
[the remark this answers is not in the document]
CalPinSW: @jamiehumphries FYI I've picked this ticket up. 

I think that makes sense. The only caveat is that we have a database table for `Epc` objects so will need to create a migration to add `IsLatestAssessmentForAddress`. Think that's fine though (although I'm a bit curious why the table is there as I can't see where it gets used/accessed in code.)

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
CalPinSW: I've had a bit of a poke around and from what I can tell, the endpoint for searching by address returns a less detailed Epc object which doesn't include the 'IsLatestAssessmentForAddress' property, whereas the Epc by id endpoint does return this property.

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1628,5 +1629,35 @@ public PathByActionArguments(string action, string controller, RouteValueDiction
                 Values = values;
             }
         }
+
+        private async Task<List<EpcSearchResult>> FilterExpiredEpc(List<EpcSearchResult> searchResults)
+        {
+            var expiredEpcIds = await GetDuplicateExpiredEpcIds(searchResults);
```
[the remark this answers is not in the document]
CalPinSW: If I'm understanding correctly, the intention of this code is:
`GetDuplicateExpiredEpcIds`:
1. Goes through the search results.
2. Finds any EPCs with duplicate addresses. 
3. Finds which of those is the most recent. 
4. Returns a string of all of those that are not the most recent (if they are part of a duplicate set).
`FilterExpiredEpc`:
1. Then this function removes EPCs with those Ids.

I'm not convinced these really need to be separate functions. I'll have a play around with the variable names and try and simplify it.

CalPinSW:
@jamiehumphries I've pushed some changes to this. If you think I've overcomplicated either the Filter function or the Api (by using and hence modifying the Epc class) let me know and I can roll back.

CalPinSW:
Also think it's worth having some tests in place for this. Currently it seems the error is not reproducible live, but apparently it's come and gone before. I'll put some in place for Monday

SeaPublicWebsite/Services/EnergyEfficiency/Epc/EpcFilterService.cs
```diff
@@ -0,0 +1,47 @@
...
+            .Where(grouping => grouping.Count() == 1)
+            .SelectMany(grouping => grouping);
+            
+        var epcSearchResultsWithAdditionalEpcInformationTaskList = epcSearchResultsNotUniqueByAddress.Select(async epc => new KeyValuePair<EpcSearchResult, BusinessLogic.Models.Epc>(epc, await epcApi.GetEpcForId(epc.EpcId)));
+
+        var epcSearchResultsWithAdditionalEpcInformation = await Task.WhenAll(epcSearchResultsWithAdditionalEpcInformationTaskList);
+
+        var epcSearchResultsNotUniqueByAddressWithOlderDuplicatesRemoved = epcSearchResultsWithAdditionalEpcInformation
+            .Where(epcDtoPair => epcDtoPair.Value.IsLatestAssessmentForAddress)
+            .Select(epcDtoPair => epcDtoPair.Key);
```
davidm-m, line 43: Might just be me not understanding how EPCs work in the codebase, but `IsLatestAssessmentForAddress` is nullable and we're not doing anything to populate it in this PR. Until the field is correctly populated for each address we risk returning nothing when there are several results.
Is that actually a risk?

jamiehumphries:
@CalPinSW, I'm going to close this PR and revert to Taiwo's original code for now. We would like to get this merged ASAP and I can't quite convince myself about all the database changes.

I think we will have to accept some potential inefficiency now. We may revisit this as tech debt.

jamiehumphries:
@CalPinSW, I'm going to close this PR and revert to Taiwo's original code for now. We would like to get this merged ASAP and I can't quite convince myself about all the database changes.

I think we will have to accept some potential inefficiency now. We may revisit this as tech debt.

SeaPublicWebsite/Services/EnergyEfficiency/Epc/EpcFilterService.cs
CalPinSW: Just in case someone is looking at this for tech debt. From how I understood it (and seemed to be the case testing locally). IsLatestAssessmentForAddress was part of the object returned by the EPC Api. 

Essentially the search by address endpoint returns a list of EPC ids with a little information on them (but not the full object). The bug is caused by this endpoint sometimes returning multiple for the same address.

The EPC by id endpoint returns a more complete object which includes `IsLatestAssessmentForAddress`. If we get 2 EPCs for an address from the search endpoint we check the full EPCs in this function. (That being said I suppose there could be a weird case where neither of the search EPCs are the latest. That might be worth thinking about).

PR #259, opened 2024-01-26
PC-572-Loft Insulation Recommendation fix

CalPinSW:
https://beisdigital.atlassian.net/browse/PC-572
- Update YearBuilt enum to include all dates from EPC API
- Update Tests 
- Update GetBreConstructionDate switch statement

PR #261, opened 2024-01-31
PC-572: Update home age options to match EPC bands

jamiehumphries:
This is a reworking of https://github.com/UKGovernmentBEIS/beis-simple-energy-advice-beta/pull/259.

The main difference is that we preserve the deprecated home age bands to support historic data.


CalPinSW (approved):
This is a really neat way of handling the deprecated enums. This all matches my understanding of the ticket. I've done a quick check locally that it works for saved Property Details using from the previous enum format. All looks good.

PR #264, opened 2024-02-23
PC-841: Fix view model binding by removing localizer

jamiehumphries:
This fixes the issue on the confirm address page investigated as PC-841.


jamiehumphries:
The view models edited were not able to bind on POST requests, which was causing the issue.

PR #293, opened 2024-06-21
Pc 1122 investigation into c# and .net update

Glenn-Clarke:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1050)

# Description

C# & .NET update changes with port number 8080 reverted to 80

# Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   sea:
     build: .
     ports:
-      - "5001:80"
+      - "5001:8080"
     environment:
       ASPNETCORE_ENVIRONMENT: "Development"
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,9 +4,10 @@ services:
   sea:
     build: .
     ports:
-      - "5001:8080"
+      - "5001:80"
     environment:
       ASPNETCORE_ENVIRONMENT: "Development"
+      ASPNETCORE_HTTP_PORTS: "80"
     volumes:
       # map the dotnet user-secret folder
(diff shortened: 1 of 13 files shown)


docker-compose.yml
```diff
@@ -7,6 +7,7 @@ services:
       - "5001:80"
     environment:
       ASPNETCORE_ENVIRONMENT: "Development"
+      ASPNETCORE_HTTP_PORTS: "80"
```
jdgage, line 10: Assuming this all works fine running locally?

jdgage (commented):
Could you have a quick look into the (I think new with .NET 8) build warnings? i.e. whether they're something that we can easily deal with now?

docker-compose.yml
Glenn-Clarke: I tested this change yesterday after you suggested trying it locally, and it appears to work fine, including PDF generation.

Glenn-Clarke:
Looks like the build warnings were simple replacements, but I'm testing the changes locally now

jdgage (dismissed):
Did you look at the warning about lower-cased characters in that migration class name? I think we should be able to just rename the class (and maybe add a comment about why)

Glenn-Clarke:
They don't seem to come up in the build... I'll have a look now quickly and see if I can find them

Glenn-Clarke:
They didn't show up in my IDE unless I went to file in question 😓 I found them using the github build warnings, and I've fixed it now. I renamed the file and class for the migrations and the data in my local db seems to be fine. Hope this is ok!

Glenn-Clarke:
Docker compose up seems to be ok? I think?

PR #344, opened 2025-01-20
PC-1431: Replace basic auth for pre-prod environments

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1431)

# Description

see [PC-1432](https://beisdigital.atlassian.net/browse/PC-1432) for original ticket. most of the code is copied from that ticket's PR

only difference to note is that some extra function was needed for the PDF generator, which uses a local chrome process. this chrome process needs to be able to bypass auth to view the page so added some code to allow it to do this

# Checklist

# Screenshots

image
image
image


README.md
```diff
@@ -197,8 +197,7 @@ The following is a list of environment variables required by the application. Ot
   - Set automatically by .NET
 - DOTNET_ENVIRONMENT
   - Set automatically by .NET   
-- BasicAuth__Password
-- BasicAuth__Username
+- Auth__Password
```
samyou-softwire, line 200: to note FWTS differs as it does _not_ set any default password. we should probably align one way or the other, could copy the HUG2 appsettings.DEV.json & appsettings.Staging.json if this is what we want to do

SeaPublicWebsite/Views/Auth/Index.cshtml
```diff
@@ -0,0 +1,37 @@
...
+@using GovUkDesignSystem
+@using GovUkDesignSystem.GovUkDesignSystemComponents
+@using SeaPublicWebsite.Controllers
+@model SeaPublicWebsite.Models.Auth.AuthViewModel
+
+@{
+    ViewBag.Title = "Input Password";
+}
+
+<div class="govuk-grid-row">
```
samyou-softwire, line 11: this page doesn't set any welsh translations. I don't think is needed for the ticket, though it does put a welsh toggle on the page which doesn't do anything. did a quick check and it doesn't seem untrivial to turn off, so left for now

SeaPublicWebsite/Middleware/AuthMiddleware.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Web;
+using Microsoft.AspNetCore.Http;
+using SeaPublicWebsite.BusinessLogic.Services.Password;
+using SeaPublicWebsite.Services;
+
+namespace SeaPublicWebsite.Middleware;
+
+public class AuthMiddleware(RequestDelegate next)
+{
+    private static readonly string[] IgnoredPaths = ["/health-check", "/password", "/compiled", "/assets"];
```
samyou-softwire, line 13: did a quick check and these paths seem to be also valid for FWTS

SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
```diff
@@ -2,22 +2,14 @@
...
-using Microsoft.Extensions.Options;
 using PuppeteerSharp;
 using PuppeteerSharp.Media;
-using SeaPublicWebsite.Middleware;
+using SeaPublicWebsite.BusinessLogic.Services.Password;
 
 namespace SeaPublicWebsite.Services.EnergyEfficiency.PdfGeneration;
 
-public class PdfGenerationService
+public class PdfGenerationService(AuthService authService, PasswordService passwordService)
```
samyou-softwire, line 11: for what its worth it was really surprising to me that we do this, especially as GBIS doesn't. could perhaps raise a ticket to investigate this behaviour further?

SeaPublicWebsite/Views/Auth/Index.cshtml
jdgage: No (by analogue there aren't any translations on the basic auth popup), although I think it's worth mentioning on the ticket.

README.md
jdgage: I think having a fallback in the codebase is a good thing, to guard against getting locked out if there's not something in AWS for whatever reason. Possibly it's confusing because you have to know it's overridden in AWS. Maybe that means it's worth a comment? But then we don't have similar comments for other settings that are overwritten.

SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
jdgage: Yeah this does feel a bit suspect, especially given there was some plan to revisit it in the past. Think it's worth raising a ticket, yes.

jdgage: Does GBIS generate PDFs? I don't think it does?

jdgage (dismissed):
Think the changes are fine, but some conversations to address

SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
samyou-softwire: ah right, I think I'm getting it mixed up with the print dialogue, which we utilise to let users save a PDF of the page. I suppose this does have the added convenience of not requiring users to use that confusing dialogue. would we consider porting this system to GBIS, or researching a better system and implementing it on both?

README.md
samyou-softwire: have added some config. I'm not sure about commeting the JSON files directly so I've added a comment to the `PasswordConfiguration` file

SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
jdgage: I think we want something better than this before we talk about porting to GBIS.

For what it's worth this would be a bit of a behaviour change too: it's generating a custom PDF rather than offering the ability to print/save the current page. Wouldn't necessarily be a straightforward port over.

samyou-softwire: right, I'll add a ticket which is to investigate this a bit further how best to do this, then we can talk about outcomes in regards to FWTS & GBIS

SeaPublicWebsite/Services/AuthService.cs
```diff
@@ -9,6 +9,7 @@ public class AuthService(IWebHostEnvironment webHostEnvironment)
 
     public bool AuthIsEnabled()
     {
+        return true;
```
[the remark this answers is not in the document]
samyou-softwire: yup, should be resolved 🙈

SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
samyou-softwire: have raised as [PC-1710](https://beisdigital.atlassian.net/browse/PC-1710)

PR #349, opened 2025-01-31
PC-1730 accessibility page regions not identified with aria landmarks 

Glenn-Clarke:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1730)

# Description

Added `<nav>` and `<header>` components surrounding the sections identified in the ticket

# Checklist

# Screenshots

The first page viewed through WAVE 
{90010F1B-07EC-4CAA-BC38-0FAF40A2E383}


SeaPublicWebsite/Views/EnergyEfficiency/ActionPlan/Partials/_ActionPlanBeforeMainLinks.cshtml
```diff
@@ -2,17 +2,19 @@
...
-                   Text = SharedLocalizer["Back"].Value,
-                   Href = Model.BackLink,
-               })
+        <nav>
+            @await Html.GovUkBackLink(new BackLinkViewModel
+            {
+                Text = SharedLocalizer["Back"].Value,
+                Href = Model.BackLink
+            })
+        </nav>
```
jdgage, line 17: I think the correct way to do this would be to make `GovUkBackLink` responsible for it. Otherwise it's something you have to remember to add everywhere. Is there a reason we can't do that?

SeaPublicWebsite/Views/Shared/_Layout.cshtml
```diff
@@ -88,13 +88,14 @@
...
-            Html = @<text>@SharedLocalizer["ThisIsANewServiceString", @ViewBag.FeedbackUrl]</text>
-        }))
+    <div class="govuk-width-container">        
+        <header>
+            @(await Html.GovUkPhaseBanner(new PhaseBannerViewModel
+            {
+                Phase = "Beta",
+                Html = @<text>@SharedLocalizer["ThisIsANewServiceString", ViewBag.FeedbackUrl]</text>
+            }))
+        </header>
```
jdgage, line 98: Likewise, I think this should be the responsibility of `GovUkPhaseBanner`. It's less of a concern because we only use this once, but I think the principle still holds.

SeaPublicWebsite/Views/EnergyEfficiency/ActionPlan/Partials/_ActionPlanBeforeMainLinks.cshtml
jdgage: Should these also have `aria-label`s?

samyou-softwire: > I think the correct way to do this would be to make `GovUkBackLink` responsible for it. Otherwise it's something you have to remember to add everywhere. Is there a reason we can't do that?

I'm not so sure; if we were to make this change the default, it could have unexpected changes on other sites. if it's not default, then it's possible to forget this config setting too

perhaps changes to the `_Layout.cshtml` or a project partial would be more appropriate?

jdgage: I think the back link should always be in a `nav` landmark, so if other sites don't have them then they'll need it. If they're already using it then they'll need to remove their own wrapper, but that's an improvement anyway. What unexpected changes are you envisaging?

(bear in mind that consumers of the library would have to update, it's not going to happen immediately)

Glenn-Clarke: I've implemented these in the gov design system and I'll push changes soon! 
I've added roles to the `<a>` tag surrounding the backlink, and the `<div>` surrounding the phase banner.

What would you expect for the aria-labels, @jdgage? 
It doesn't appear that aria tags are used a lot in the partials I've seen in GovDesignLibrary, so I don't have a good template to use

jdgage: There's some examples [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role). Maybe it's not necessary here as there's only one link in the group. Will leave that to you.

On your changes, is there a particular reason you've gone for roles rather than dedicated `nav`/`header` elements? I think the latter are more semantic and preferred if possible (mentioned in that same link above).

Glenn-Clarke: I think I was preferring them to avoid having "empty" elements, in case they interfered with the CSS somehow, but I'm happy to make them into the full elements since they seem to be preferred.

samyou-softwire (commented):
at a quick scan of the HTML I don't see this aria label, am I missing it?
image

samyou-softwire (approved):
was using an outdated local copy. I see the labels now
image

PR #372, opened 2025-03-03
PC-1760: Add container user

Glenn-Clarke:
Small fix to specify container user

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,8 +37,7 @@ RUN dotnet publish -c $CONFIGURATION -o out
 
 # Build runtime image
 FROM base
-RUN groupadd -r appuser && useradd -r -g appuser appuser
-USER appuser
+USER app
 
 WORKDIR /SeaPublicWebsite
 COPY --from=build-env /SeaPublicWebsite/out .
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,5 +42,5 @@ USER appuser
 
 WORKDIR /SeaPublicWebsite
 COPY --from=build-env /SeaPublicWebsite/out .
-EXPOSE 80
+EXPOSE 8080
 ENTRYPOINT ["dotnet", "SeaPublicWebsite.dll"]
--- a/SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
+++ b/SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
@@ -51,6 +51,6 @@ public class PdfGenerationService(AuthService authService, PasswordService passw
     private string GetLocalAddress()
     {
         // If the port the application runs on is ever changed this will need to be updated
-        return "http://localhost:80";
+        return "http://localhost:8080";
     }
 }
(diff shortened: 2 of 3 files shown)


SeaPublicWebsite.UnitTests/SeaPublicWebsite.UnitTests.csproj
```diff
@@ -5,7 +5,8 @@
         <IsPackable>false</IsPackable>
         <RootNamespace>Tests</RootNamespace>
         <IsPublishable>false</IsPublishable> 
-        <LangVersion>12</LangVersion> 
+        <LangVersion>12</LangVersion>
```
samyou-softwire, line 8: ?

Glenn-Clarke: I think it's a trailing space that I deleted, potentially?

Glenn-Clarke: it appears to be the case, in my latest commit I did the same for the line above (as no trailing space seems to be the standard) and it's got the same diff as the one you're seeing here.

samyou-softwire (approved):
LGTM

I suppose perhaps a question of can we change the container user to be the one ASP.NET is expecting so we don't have to manually override like this, though feels more reliable to specify this explicitly

Glenn-Clarke:
I think playing this one like we did with the port numbers (setting what we're using ourselves, then using it ourselves) is the more reliable approach. 
This way we're only relying on ourselves, but if we rely on ASP.NET not changing things then it's one more thing to go wrong.

jdgage:
So as discussed, I don't think this is the issue here: I expect it's that the non-root user doesn't have permission to expose the app on port 80. So we should use 8080 instead, and get the ECS task definition updated with the new mapping. Most of the changes are going to sit with ICS Digital: the task definition and a change to the pipeline as well (as it's there that it's actually trying to expose on port 80).

Something else I found here is [this](https://learn.microsoft.com/en-us/dotnet/core/compatibility/containers/8.0/app-user), which we could use rather than creating our own user.

Glenn-Clarke:
I've made a small update here now instead, since we're changing the local port

Glenn-Clarke:
> You're also going to want to update the `EXPOSE` instruction in the Dockerfile, as well as the references to the ports in the `docker-compose.yml`.
> 
> See also the end of the comment above about the `app` user: we may as well take the opportunity to switch to that.

Oops, they didn't show up during my text search 🤔  working on it now

Glenn-Clarke:
For the 'app' user change, I'll push my changes now, but it seems like that user definitely *exists* but can't access the secrets (api keys) so the application crashes due to null reference exceptions.

jdgage:
> For the 'app' user change, I'll push my changes now, but it seems like that user definitely _exists_ but can't access the secrets (api keys) so the application crashes due to null reference exceptions.

Not sure I follow: when's it trying to do that?

Glenn-Clarke:
Missed a reference to the user in a path! Sorting it now.

Glenn-Clarke:
@jdgage It appears the necessary changes have been made to the dev environment, it is currently back up and running. 
Would you mind reviewing this again so I can let the Cloud Operations Team know if the dev environment changes are also working with these additional changes?

jdgage (approved):

PR #347, opened 2025-01-30
PC-1710: Improve PDF generating method

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1710)

# Description

in order to use sandboxed mode, we need docker to use a non root user

container is changed to now use a non root user. our external state uses the database which shouldn't be affected by a user change.

# Checklist

- [x] I have made any necessary updates to the documentation
- [ ] If necessary, I've added QA guidance notes to the original ticket
- [x] I have checked there are no unnecessary IDE warnings in this PR
- [x] I have checked there are no unintentional line ending changes
- [x] I have added tests where applicable
- [x] If I have made any changes to the code, I have used the IDE auto-formatter on it

PR #374, opened 2025-03-11
PC-1710 Improve PDF generation method

Glenn-Clarke:
Added --no-sandbox flag as a test


SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Globalization;
+﻿using System;
```
samyou-softwire, line 1: is this needed?

samyou-softwire (approved):
looks to be the reverse of https://github.com/UKGovernmentBEIS/beis-simple-energy-advice-beta/pull/347/files. happy to give this a go to see if it resolves the issue

PR #376, opened 2025-03-13
PC-1686: Update Entity Framework

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1686)

# Description

bumps all EntityFramework packages & npgsql

dependabot raised them separately before as the package compatibility means they may be updated independently, but in practice this causes the site to not run

see #352 & #336 

7.0 upgrade requires we can no longer use [UseXMinAsConcurrencyToken()](https://www.npgsql.org/efcore/release-notes/7.0.html#obsoleted-usexminasconcurrencytoken). have switched to the solution they recommend

8.0 upgrade requires we [no longer specify Integrated Security in the connection string](https://www.npgsql.org/doc/release-notes/8.0.html#obsoletions-and-obsolete-api-removals). **this will need ICS coordination to release, leaving integrated security in the connection string will cause a db connection fail**

update: documentation says that the string can be removed without consequence in current versions of FWTS. testing locally this does appear to be true, so we can ask ICS to make the changes ahead of time and it should deploy fine

# Checklist


SeaPublicWebsite/SeaPublicWebsite.csproj
```diff
@@ -1,93 +1,93 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
```
samyou-softwire, line 1: there looks to be a line endings change here, though my IDE reports all my csproj files as CRLF. this true for everyone else?

jdgage:
> 8.0 upgrade requires we [no longer specify Integrated Security in the connection string](https://www.npgsql.org/doc/release-notes/8.0.html#obsoletions-and-obsolete-api-removals). this will need ICS coordination to release, leaving integrated security in the connection string will cause a db connection fail
> 
> update: documentation says that the string can be removed without consequence in current versions of FWTS. testing locally this does appear to be true, so we can ask ICS to make the changes ahead of time and it should deploy fine

Let's get ICS Digital to do this before merging this, so we can test it's all fine without the upgrade. Then we can go ahead with the other environments before this gets released.

SeaPublicWebsite/SeaPublicWebsite.csproj
jdgage: Same for me. You should be able to check the files before and after explicitly (open in Notepad++ or something) to see what characters are there.

SeaPublicWebsite.Data/SeaDbContext.cs
```diff
@@ -35,7 +35,9 @@ private void SetupPropertyData(ModelBuilder modelBuilder)
             .HasKey("PropertyDataId");
         
         // Property data row versioning
-        AddRowVersionColumn(modelBuilder.Entity<PropertyData>());
+        modelBuilder.Entity<PropertyData>()
+            .Property(p => p.Version)
+            .IsRowVersion();
```
jdgage, line 40: There's part of me that thinks we should have an `IEntityWithRowVersioning` interface that enforces a `uint Version` property and is implemented by these three classes. Then our models enforce the versioning a bit explicitly, and we can still have the common method for adding this, with signature:
```
private void AddRowVersionColumn<T>(EntityTypeBuilder<T> builder) where T : IEntityWithRowVersioning
```

What do you think?

SeaPublicWebsite.BusinessLogic/Models/Epc.cs
```diff
@@ -21,6 +21,7 @@ public class Epc
     public GlazingType? GlazingType { get; set; }
     public HasHotWaterCylinder? HasHotWaterCylinder { get; set; }
     public SolarElectricPanels? HasSolarElectricPanels { get; set; }
+    public uint Version { get; set; }
```
jdgage, line 24: Checking: these columns were already secretly there, so there's no migration to add them?

SeaPublicWebsite/SeaPublicWebsite.csproj
samyou-softwire: inspected in the github online editor and the file was changed to LF, not sure where they came from but have pushed a fixup to make it CRLF again

SeaPublicWebsite.BusinessLogic/Models/Epc.cs
samyou-softwire: yeah when generating a migration it comes up empty

though the snapshot does change so will push an update to that

SeaPublicWebsite.Data/SeaDbContext.cs
samyou-softwire: sounds good, suppose it does make me wonder a little why this isn't exposed by the library. think this is a bit closer to what GBIS does as well

have made the change

SeaPublicWebsite.Data/SeaDbContext.cs
```diff
@@ -93,9 +93,11 @@ private void SetupRelations(ModelBuilder modelBuilder)
             .OnDelete(DeleteBehavior.Cascade);
     }
 
-    private void AddRowVersionColumn<T>(EntityTypeBuilder<T> builder) where T : class
+    private void AddRowVersionColumn<T>(EntityTypeBuilder<T> builder) where T : class, IEntityWithRowVersioning
     {
-        // This is a PostgreSQL specific implementation of row versioning
-        builder.UseXminAsConcurrencyToken();
+        // Epc row versioning
```
jdgage, line 98: Don't want this

SeaPublicWebsite.Data/Migrations/SeaDbContextModelSnapshot.cs
```diff
@@ -100,13 +100,14 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<int?>("SolidWallsInsulated")
                         .HasColumnType("integer");
 
-                    b.Property<int?>("WallConstruction")
-                        .HasColumnType("integer");
-
-                    b.Property<uint>("xmin")
```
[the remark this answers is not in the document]
jdgage: Oh yes you're right, my mistake!

SeaPublicWebsite.Data/SeaDbContext.cs
```diff
@@ -93,9 +93,11 @@ private void SetupRelations(ModelBuilder modelBuilder)
             .OnDelete(DeleteBehavior.Cascade);
     }
 
-    private void AddRowVersionColumn<T>(EntityTypeBuilder<T> builder) where T : class
+    private void AddRowVersionColumn<T>(EntityTypeBuilder<T> builder) where T : class, IEntityWithRowVersioning
     {
-        // This is a PostgreSQL specific implementation of row versioning
-        builder.UseXminAsConcurrencyToken();
+        // Use database specific implementation of row versioning
```
jdgage, line 98: I think mentioning Postgres was right here. But tbh I think the only thing the comment should do is link to the docs that suggest this.

samyou-softwire: have done so

samyou-softwire:
ICS have made the change and dev still seems to work so merge should no longer be blocked

PR #377, opened 2025-03-13
PC-1686: Reduce dependabot scan frequency to monthly

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1686)

# Description

ticket initially requested weekly, though as we have since agreed will not check these very often (bi monthly) it helps to set the interval large as it will group all updates in a month together

# Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: "nuget"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
     # we may not be able to respond to pull requests immediately
     # prioritise showing them all
     open-pull-requests-limit: 100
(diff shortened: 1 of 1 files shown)


jdgage (dismissed):

jdgage:
Does Dependabot not update the existing PRs with any new changes where it's grouping them together?

jdgage:
I think weekly would be better even if we're doing it every couple of months. Depending on timing, there's a risk that we lose out on nearly a month's worth of updates.

samyou-softwire:
> Does Dependabot not update the existing PRs with any new changes where it's grouping them together?

looks to be that it will raise a new PR at each interval

> I think weekly would be better even if we're doing it every couple of months. Depending on timing, there's a risk that we lose out on nearly a month's worth of updates.

I think we expect that two updates will have run between each ticket. especially as FWTS has few dependencies we can except the potential amount of updates to miss will be quite small anyway

for reference any updates linked to security issues will be raised immediately and will always be caught

samyou-softwire:
> > looks to be that it will raise a new PR at each interval
> 
> And close the old one?

looks like it'll close the old one

jdgage:
If it's creating new ones and closing the old ones, then I think we may as well have it run more frequently

samyou-softwire:
have changed to weekly

PR #391, opened 2025-03-17
Minor tweaks

jdgage:


SeaPublicWebsite/Controllers/HomeController.cs
```diff
@@ -7,10 +7,10 @@ public class HomeController : Controller
     [HttpGet("/")]
     public IActionResult Index()
     {
+#pragma warning disable CS0162 // Unreachable code detected
```
samyou-softwire, line 10: out of interest was there a warning around this, or is this to make the section a little more correct?

samyou-softwire (approved):
LGTM

SeaPublicWebsite/Controllers/HomeController.cs
jdgage: If you changed your build configuration in Rider, then it would (correctly) flag the other branch as unreachable instead.

PR #395, opened 2025-03-18
PC-1760: --no-sandbox Documentation & Investigation Results

Glenn-Clarke:
Added documentation to the codebase referencing the findings from this ticket's investigation.

As part of this investigation I've also verified that the PDF generator uses localhost:8080+path, so it should only be able to generate PDFs of items within the app itself.

### DIFF
--- a/SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
+++ b/SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
@@ -17,6 +17,10 @@ public class PdfGenerationService(AuthService authService, PasswordService passw
             Headless = true,
             Args =
             [
+                /* Investigated this flag's removal in PC-1760
+                // Due to limitations with running chromium sandboxes in Fargate, we are unable to remove this flag
+                // See GitHub issue: https://github.com/aws/containers-roadmap/issues/2102
+                */ 
                 "--no-sandbox"
             ]
         };


samyou-softwire (approved):
LGTM

SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
```diff
@@ -17,6 +17,10 @@ public async Task<Stream> GeneratePdf(string path)
             Headless = true,
             Args =
             [
+                /* Investigated this flag's removal in PC-1760
+                // Due to limitations with running chromium sandboxes in Fargate, we are unable to remove this flag
+                // See GitHub issue: https://github.com/aws/containers-roadmap/issues/2102
+                */
```
jdgage, line 23: I think it's better to summarise the conclusion here rather than refer to the ticket:
1. There's no guarantee that someone reading this comment has access to that Jira project
2. They can always get the ticket number from the commit message if they want it

So I think this needs to mention the conclusions we found:
* The `--no-sandbox` flag is generally considered bad practice, as it means anything running in the browser could have root access to anything else in the container
* ECS tasks on Fargate don't currently support sandboxing, so we're forced to use this flag
* We only serve our content on this browser process, and we trust that*, so we're not concerned in the interim
* If trying to use this method for anything else, you should bear all this in mind and ensure we're only serving trusted content

*And this is the important thing to check. Note that it's not sufficient that it's only going to `localhost:8080`: if there was a way to inject user-generated content into the page it renders then that would be a way in (you could perhaps inject a `<script>` that the browser would then execute and would have permission to escape into the container). What determines the content on that page? Is there anything that directly comes from the user?

jdgage (changes requested):

SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
Glenn-Clarke: Good point, I'll revise my comment and review if we serve anything user-entered.

Glenn-Clarke: I've had a look, and the only user entered thing that could be something is that the POST request to generate the PDF has /{reference} at the end, this reference is then taken and used in the URL of the page we GET to generate the PDF.

I've had a look online and I don't think this is an issue, as any script would have to start the url with "javascript:" which we explicitly do not allow by appending to localhost:8080. 

I would like your input before I say for certain on this though @jdgage

jdgage: Do we do any validation on the reference in the URL? If someone puts something else instead of their reference code and we `GET localhost:8080/energy-efficiency/pdf-generation/your-recommendations/<something else>`, what happens?

To illustrate, the sort of scenario I'm talking about:
1. There's an input on the site somewhere that lets the user enter their name, say
2. We take that value the user's given us and later render it on this page, without escaping it

Then a user could come along and input their name as `<script>alert('oh no!')</script>` and that script would get injected into the page and run in the headless browser. And that script could be anything malicious.

The analogue here would be if you could `POST /energy-efficiency/pdf-generation/your-recommendations/<script>bad things</script>` and it would run that malicious script in the browser. Hopefully we're doing various things that make that impossible, for example stopping users POSTing any old thing there, or validating that it's actually a reference code in the path, or HTML-escaping that parameter, or not rendering it on the page at all!

Does that make sense?

Glenn-Clarke: Hi Joe, yes, I think my previous message was a bit unclear regarding what I checked. I checked that when creating a PDF, no user input strings etc. can be viewed, only things defined by enums which the user has selected. In PDF mode (_ReturnToActionPlan.cshtml), the email input box on the page is disabled as well.

The string in the URL being used in the "GET" for the request to the PDF generation was the thing I was unsure about.
We take whatever string is submitted in that post request, and then URL Encode it, and then append that verbatim to the path we query to generate the PDF.
```c#
        [HttpPost("action-plan-download/{reference}")]
        public async Task<IActionResult> GenerateActionPlanPdf_Post(string reference)
        {
            var encodedReference = HttpUtility.UrlEncode(reference);
            var stream = await pdfGenerationService.GeneratePdf($"energy-efficiency/pdf-generation/action-plan/{encodedReference}");
            return File(stream, MediaTypeNames.Application.Pdf, "ActionPlan.pdf");
        }
``` 
We query the database with this reference to retrieve the user's information on the headless browser, so I think EFCore should prevent injection attacks, and we'd return an error if an invalid page was requested.

jdgage: You're right, but I think it's more of a side effect (we're not explicitly validating the reference code). More important is that, from a quick look, we don't actually use the given reference code in rendering the page at all - we take it from the database, and what's in the database is something we've generated ourselves, not user input.

SeaPublicWebsite/Services/EnergyEfficiency/PdfGeneration/PdfGenerationService.cs
```diff
@@ -10,13 +10,24 @@ namespace SeaPublicWebsite.Services.EnergyEfficiency.PdfGeneration;
...
         var launchOptions = new LaunchOptions
         {
             Headless = true,
             Args =
             [
+                /*
+                //'--no-sandbox' is not considered good practice as what is running in this browser can have root access to the container
+                // This is only used here as ECS Tasks on Fargate do not currently support sandboxing
+                // See GitHub issue: https://github.com/aws/containers-roadmap/issues/2102
+                // At present, we aren't concerned about this as we only serve our own content on this browser process
```
jdgage, line 28: and we don't render unvalidated user input

jdgage (approved):

PR #402, opened 2025-03-31
PC-1500: Heating Control EPC Integration

Glenn-Clarke:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1500)

# Description

Added EPC Hint support to the Heating Controls question.

If no EPC information supplied -> Hint does not show
If information is supplied but none is relevant, or array is empty -> Hint shows suggesting "No heating controls"
If any of the heating controls are present -> Hint shows with relevant controls

**Review Note:**
I've also included a small edit to an old migration, this should represent no change in actual functionality, but it appears the old way of setting a default value for arrays no longer works since the dependency upgrades, and results in IndexOutOfRange exceptions when trying to build the project locally.

As the migration is already in place on live, I don't expect this to cause any issues because the migration shouldn't be reapplied? And even if it were to be reapplied, it should represent no change in actual functionality, it is an alternate way of setting the default value to an empty array.

Please take a good look at this migration when reviewing.

# Checklist

# Screenshots

{1AA177AB-A3E7-42AC-BF1C-58D0ABA52A50}

{E3F5700E-607C-44E4-A014-1942400954F7}

{5E426E5E-3DA3-4B9A-93B5-1B6CDD0AC4DB}
{38BBF4F5-B0E9-4D59-9C28-0781E6EF80DF}

{8B0FC7FF-C7FA-412B-8AF0-44680DFE93CD}

{AC7823CA-8285-4534-AF35-D7EDF512846C}

{F98CB1B9-A56B-492D-B67D-AA8A9DACAAA9}

{97CC8EC9-CDD7-4E7D-9083-767B4FDE5C94}

### DIFF
--- a/SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
+++ b/SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
@@ -67,6 +67,9 @@ public class EpbEpcAssessmentDto
     [JsonProperty(PropertyName = "photoSupply")]
     public int? SolarElectricPanelPercentRoofCoverage { get; set; }
 
+    [JsonProperty(PropertyName = "mainHeatingControls")]
+    public string[] MainHeatingControls { get; set; }
+
     public Epc Parse()
     {
         if (AssessmentType.Equals("SAP", StringComparison.OrdinalIgnoreCase))
@@ -92,7 +95,8 @@ public class EpbEpcAssessmentDto
             GlazingType = ParseGlazingType(),
             EpcHeatingType = ParseHeatingType(),
             HasHotWaterCylinder = ParseHasHotWaterCylinder(),
-            HasSolarElectricPanels = ParseSolarElectricPanelPercentRoofCoverage()
+            HasSolarElectricPanels = ParseSolarElectricPanelPercentRoofCoverage(),
+            HeatingControls = ParseMainHeatingControls()
         };
     }
 
@@ -776,5 +780,26 @@ public class EpbEpcAssessmentDto
         return SolarElectricPanelPercentRoofCoverage > 0
             ? SolarElectricPanels.Yes
             : SolarElectricPanels.No;
+    }
+
+    private List<HeatingControls> ParseMainHeatingControls()
+    {
+        if (MainHeatingControls == null) return [];
+
+        List<HeatingControls> output = [];
+        if (MainHeatingControls.Any(s => s.Contains("Programmer", StringComparison.OrdinalIgnoreCase)))
+        {
+            output.Add(HeatingControls.Programmer);
+        }
+        if (MainHeatingControls.Any(s => s.Contains("Room thermostat", StringComparison.OrdinalIgnoreCase)))
+        {
+            output.Add(HeatingControls.RoomThermostats);
+        }
+        if (MainHeatingControls.Any(s => s.Contains("TRV") || s.Contains("Thermostatic Radiator Valve", StringComparison.OrdinalIgnoreCase)))
+        {
+            output.Add(HeatingControls.ThermostaticRadiatorValves);
+        }
+        
+        return output.Count > 0 ? output : [HeatingControls.None];
     }
 }
--- a/SeaPublicWebsite.Data/Migrations/20250113163718_AddHeatingControlsField.cs
+++ b/SeaPublicWebsite.Data/Migrations/20250113163718_AddHeatingControlsField.cs
@@ -13,7 +13,7 @@ namespace SeaPublicWebsite.Data.Migrations
                 name: "HeatingControls",
                 table: "PropertyData",
                 type: "integer[]",
-                defaultValue: "{}",
+                defaultValueSql: "ARRAY[]::integer[]",
                 nullable: false);
         }
 
--- a/SeaPublicWebsite.UnitTests/EpcTests.cs
+++ b/SeaPublicWebsite.UnitTests/EpcTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using FluentAssertions;
 using NUnit.Framework;
@@ -573,6 +574,45 @@ public class EpcTests
                 }))
             .ToArray();
 
+    private static readonly EpcTestCase[] HeatingControlsTestCases =
+        new (string Descrption, string[] inputHeatingControls, List<HeatingControls> outputHeatingControls)[]
+            {
+                ("Can parse null", null, []),
+                ("Can parse no heating controls (empty array)", [], [HeatingControls.None]),
+                ("Can parse no heating controls (no relevant entries)", ["fake heating control"],
+                    [HeatingControls.None]),
+                ("Can parse Room Thermostats", ["Room Thermostats"], [HeatingControls.RoomThermostats]),
+                ("Can parse Programmer", ["Programmer"], [HeatingControls.Programmer]),
+                ("Can parse Thermostatic Radiator Valves (TRVs)", ["Thermostatic Radiator Valves"],
+                    [HeatingControls.ThermostaticRadiatorValves]),
+                ("Can parse Programmer, Room Thermostats", ["Programmer, Room Thermostats"],
+                    [HeatingControls.Programmer, HeatingControls.RoomThermostats]),
+                ("Can parse Room Thermostats, Thermostatic Radiator Valves (TRVs)",
+                    ["Thermostatic Radiator Valves (TRVs), Room Thermostats"],
+                    [HeatingControls.RoomThermostats, HeatingControls.ThermostaticRadiatorValves]),
+                ("Can parse Programmer, Thermostatic Radiator Valves (TRVs)",
+                    ["Programmer, Thermostatic Radiator Valves (TRVs)"],
+                    [HeatingControls.Programmer, HeatingControls.ThermostaticRadiatorValves]),
+                ("Can parse Programmer, Room Thermostats, Thermostatic Radiator Valves (TRVs)",
+                    ["Programmer, Room Thermostats, Thermostatic Radiator Valves (TRVs)"],
+                    [
+                        HeatingControls.Programmer, HeatingControls.RoomThermostats,
+                        HeatingControls.ThermostaticRadiatorValves
+                    ])
+            }
+            .Select(p => new EpcTestCase(
+                p.Descrption,
+                new EpbEpcAssessmentDto
+                {
+                    AssessmentType = "RdSAP",
+                    MainHeatingControls = p.inputHeatingControls
+                },
+                new Epc
+                {
+                    HeatingControls = p.outputHeatingControls
+                }))
+            .ToArray();
+
     private static readonly EpcTestCase[] EpcParseTestCases =
         Array.Empty<EpcTestCase>()
             .Concat(AssessmentTypeTestCases)
@@ -593,6 +633,7 @@ public class EpcTests
             .Concat(HeatingTypeTestCases)
             .Concat(HasHotWaterCylinderTestCases)
             .Concat(SolarElectricPanelsTestCases)
+            .Concat(HeatingControlsTestCases)
             .ToArray();
 
     public class EpcTestCase
--- a/SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
+++ b/SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
@@ -1082,6 +1082,7 @@ namespace SeaPublicWebsite.Controllers
             {
                 HeatingControls = propertyData.HeatingControls,
                 Reference = propertyData.Reference,
+                Epc = propertyData.Epc,
                 EntryPoint = entryPoint,
                 BackLink = GetBackUrl(QuestionFlowStep.HeatingControls, propertyData, entryPoint)
             };
--- a/SeaPublicWebsite/Helpers/EnumDisplayExtensions.cs
+++ b/SeaPublicWebsite/Helpers/EnumDisplayExtensions.cs
@@ -1,4 +1,5 @@
-﻿using GovUkDesignSystem.Attributes;
+﻿using System;
+using GovUkDesignSystem.Attributes;
 using SeaPublicWebsite.BusinessLogic.Models.Enums;
 
 namespace SeaPublicWebsite.Helpers;
@@ -10,5 +11,16 @@ public static class EnumDisplayExtensions
         return floorConstruction is FloorConstruction.Mix
             ? "A mix of both suspended timber and solid concrete"
             : GovUkRadioCheckboxLabelTextAttribute.GetLabelText(floorConstruction);
+    }
+
+    public static string HeatingControlEnumToEpcHintResourceKey(this HeatingControls heatingControl)
+    {
+        return heatingControl switch
+        {
+            HeatingControls.Programmer => "ProgrammerHeatingControlEPCHintString",
+            HeatingControls.RoomThermostats => "RoomThermostatHeatingControlEPCHintString",
+            HeatingControls.ThermostaticRadiatorValves => "ThermostaticRadiatorValvesHeatingControlEPCHintString",
+            _ => throw new IndexOutOfRangeException()
+        };
     }
 }
--- a/SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
+++ b/SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
@@ -4,6 +4,8 @@
 @using SeaPublicWebsite.BusinessLogic.Models.Enums
 @using SeaPublicWebsite.Controllers
 @using Microsoft.AspNetCore.Mvc.Localization
+@using Notify.Exceptions
+@using SeaPublicWebsite.Helpers
 @inject IHtmlLocalizer<SharedResources> SharedLocalizer
 
 @model SeaPublicWebsite.Models.EnergyEfficiency.HeatingControlsViewModel;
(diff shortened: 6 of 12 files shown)


Glenn-Clarke:
Thanks Sammy! I forgot we could do that now :)

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
```diff
@@ -777,4 +781,25 @@ private static bool HasSolidWalls(string description)
...
+        }
+        if (MainHeatingControls.Any(s => s.Contains("Room thermostat", StringComparison.OrdinalIgnoreCase)))
+        {
+            output.Add(HeatingControls.RoomThermostats);
+        }
+        if (MainHeatingControls.Any(s => s.Contains("TRV") || s.Contains("Thermostatic Radiator Valve", StringComparison.OrdinalIgnoreCase)))
+        {
+            output.Add(HeatingControls.ThermostaticRadiatorValves);
+        }
+
```
samyou-softwire, line 802: here

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1082,6 +1082,7 @@ public async Task<IActionResult> HeatingControls_Get(string reference, QuestionF
             {
                 HeatingControls = propertyData.HeatingControls,
                 Reference = propertyData.Reference,
+                Epc = propertyData.Epc,
```
samyou-softwire, line 1085: here

samyou-softwire (changes requested):
there's a couple files with pending formatting tweaks

Glenn-Clarke:
I've run the auto formatter on those files now, I must have missed them as I went along 😅

SeaPublicWebsite/Helpers/EnumDisplayExtensions.cs
```diff
@@ -11,4 +12,15 @@ public static string DescriptionForAnswerSummary(this FloorConstruction? floorCo
...
     }
+
+    public static string HeatingControlEnumToEpcHintResourceKey(this HeatingControls heatingControl)
+    {
+        return heatingControl switch
+        {
+            HeatingControls.Programmer => "ProgrammerHeatingControlEPCHintString",
+            HeatingControls.RoomThermostats => "RoomThermostatHeatingControlEPCHintString",
+            HeatingControls.ThermostaticRadiatorValves => "ThermostaticRadiatorValvesHeatingControlEPCHintString",
+            _ => throw new IndexOutOfRangeException()
```
samyou-softwire, line 23: for both these exceptions I'm not sure if `IndexOutOfRangeException()` is the most meaningful. for most other cases with enums we use `ArgumentOutOfRangeException()`

SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
```diff
@@ -33,6 +35,29 @@
...
+                                            SharedLocalizer[Model.Epc?.HeatingControls[0].HeatingControlEnumToEpcHintResourceKey()]],
+                                    2 => SharedLocalizer["TwoHeatingControlsEPCHintString", Model.Epc.LodgementYear,
+                                        SharedLocalizer[Model.Epc?.HeatingControls[0].HeatingControlEnumToEpcHintResourceKey()],
+                                        SharedLocalizer[Model.Epc?.HeatingControls[1].HeatingControlEnumToEpcHintResourceKey()]],
+                                    3 => SharedLocalizer["ThreeHeatingControlsEPCHintString", Model.Epc.LodgementYear,
+                                        SharedLocalizer[Model.Epc?.HeatingControls[0].HeatingControlEnumToEpcHintResourceKey()],
+                                        SharedLocalizer[Model.Epc?.HeatingControls[1].HeatingControlEnumToEpcHintResourceKey()],
+                                        SharedLocalizer[Model.Epc?.HeatingControls[2].HeatingControlEnumToEpcHintResourceKey()]],
+                                    _ => throw new IndexOutOfRangeException()
```
samyou-softwire, line 56: this one could be more generic as well. it's technically index related but the index is outside the bounds of meaning, not outside the bounds of an array as this exception implies

samyou-softwire (changes requested):

SeaPublicWebsite/Helpers/EnumDisplayExtensions.cs
Glenn-Clarke: Agreed! I think I had IndexOutOfRange in mind from the migration issue at the time.

SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
Glenn-Clarke: I'll swap this for an argument out of range exception also

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
```diff
@@ -777,4 +781,25 @@ private static bool HasSolidWalls(string description)
             ? SolarElectricPanels.Yes
             : SolarElectricPanels.No;
     }
+
+    private List<HeatingControls> ParseMainHeatingControls()
```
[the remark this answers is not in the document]
Glenn-Clarke: We have this example from the maintainers: 
["Programmer, room thermostat and TRVs", "Time and temperature zone control"]

And the documentation states: "Type of main heating controls. Includes both main heating systems if there are two."

samyou-softwire:
for reference we can use the same reference code we used last time to check compatibility MCNM8LV7

SeaPublicWebsite.Data/Migrations/20250113163718_AddHeatingControlsField.cs
```diff
@@ -13,7 +13,7 @@ protected override void Up(MigrationBuilder migrationBuilder)
                 name: "HeatingControls",
                 table: "PropertyData",
                 type: "integer[]",
-                defaultValue: "{}",
+                defaultValueSql: "ARRAY[]::integer[]",
```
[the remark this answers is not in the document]
Glenn-Clarke: https://www.postgresql.org/docs/current/sql-expressions.html#:~:text=You%20can%20construct%20an%20empty%20array%2C%20but%20since%20it%27s%20impossible%20to%20have%20an%20array%20with%20no%20type%2C%20you%20must%20explicitly%20cast%20your%20empty%20array%20to%20the%20desired%20type.%20For%20example%3A

This shows that we have to explicitly type our empty arrays in this way

SeaPublicWebsite.UnitTests/EpcTests.cs
```diff
@@ -573,6 +574,45 @@ public void CanParseEpcDto(EpcTestCase testCase)
                 }))
             .ToArray();
 
+    private static readonly EpcTestCase[] HeatingControlsTestCases =
+        new (string Descrption, string[] inputHeatingControls, List<HeatingControls> outputHeatingControls)[]
+            {
+                ("Can parse null", null, []),
```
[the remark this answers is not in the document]
Glenn-Clarke: Will do

samyou-softwire (dismissed):

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
```diff
@@ -777,4 +781,28 @@ private static bool HasSolidWalls(string description)
...
+    {
+        if (MainHeatingControls == null) return [];
+
+        List<HeatingControls> output = [];
```
jdgage, line 789: Resharper used to prefer something like `var output = new List<HeatingControls>();` for this. Has that changed in newer language versions?

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
```diff
@@ -777,4 +781,28 @@ private static bool HasSolidWalls(string description)
...
+        if (MainHeatingControls.Any(s => s.Contains("Programmer", StringComparison.OrdinalIgnoreCase)))
```
jdgage, line 790: Think a newline here would be good

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
```diff
@@ -777,4 +781,28 @@ private static bool HasSolidWalls(string description)
             ? SolarElectricPanels.Yes
             : SolarElectricPanels.No;
     }
+
+    private List<HeatingControls> ParseMainHeatingControls()
+    {
+        if (MainHeatingControls == null) return [];
```
jdgage, line 787: Empty list rather than `[HeatingControls.None]`?

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
```diff
@@ -777,4 +781,28 @@ private static bool HasSolidWalls(string description)
             ? SolarElectricPanels.Yes
             : SolarElectricPanels.No;
     }
+
+    private List<HeatingControls> ParseMainHeatingControls()
```
jdgage, line 785: Does any of the logic here want to change given the latest values Temi's put on the ticket?

Whatever we go for, the really important thing will be to clearly document what we've implemented on the ticket (and get DESNZ to say they're happy with it), so it's there to refer back to.

SeaPublicWebsite.Data/Migrations/20250113163718_AddHeatingControlsField.cs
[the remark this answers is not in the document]
jdgage: Did you have to manually change this? Is this a problem with EF, or something we shouldn't be doing?

SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
```diff
@@ -33,6 +34,29 @@
...
+                                            SharedLocalizer[Model.Epc?.HeatingControls[0].HeatingControlEnumToEpcHintResourceKey()]],
+                                    2 => SharedLocalizer["TwoHeatingControlsEPCHintString", Model.Epc.LodgementYear,
+                                        SharedLocalizer[Model.Epc?.HeatingControls[0].HeatingControlEnumToEpcHintResourceKey()],
+                                        SharedLocalizer[Model.Epc?.HeatingControls[1].HeatingControlEnumToEpcHintResourceKey()]],
+                                    3 => SharedLocalizer["ThreeHeatingControlsEPCHintString", Model.Epc.LodgementYear,
+                                        SharedLocalizer[Model.Epc?.HeatingControls[0].HeatingControlEnumToEpcHintResourceKey()],
+                                        SharedLocalizer[Model.Epc?.HeatingControls[1].HeatingControlEnumToEpcHintResourceKey()],
+                                        SharedLocalizer[Model.Epc?.HeatingControls[2].HeatingControlEnumToEpcHintResourceKey()]],
+                                    _ => throw new ArgumentOutOfRangeException()
```
jdgage, line 55: When would this happen? It would be useful to give it an error message.

SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
```diff
@@ -33,6 +34,29 @@
...
+                if (Model.Epc?.HeatingControls?.Count > 0)
+                {
+                    blueBannerHint =
+                        @<p class="govuk-body">
+                            @(Model.Epc?.HeatingControls.Count switch
+                                {
+                                    1 => Model.Epc.HeatingControls[0] == HeatingControls.None
+                                        ? SharedLocalizer["NoHeatingControlsEPCHintString", Model.Epc.LodgementYear]
+                                        : SharedLocalizer["OneHeatingControlEPCHintString", Model.Epc.LodgementYear,
+                                            SharedLocalizer[Model.Epc?.HeatingControls[0].HeatingControlEnumToEpcHintResourceKey()]],
```
jdgage, line 47: Why do we sometimes have the null-check and sometimes not?

SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
```diff
@@ -33,6 +34,29 @@
...
+                                    2 => SharedLocalizer["TwoHeatingControlsEPCHintString", Model.Epc.LodgementYear,
+                                        SharedLocalizer[Model.Epc?.HeatingControls[0].HeatingControlEnumToEpcHintResourceKey()],
+                                        SharedLocalizer[Model.Epc?.HeatingControls[1].HeatingControlEnumToEpcHintResourceKey()]],
+                                    3 => SharedLocalizer["ThreeHeatingControlsEPCHintString", Model.Epc.LodgementYear,
+                                        SharedLocalizer[Model.Epc?.HeatingControls[0].HeatingControlEnumToEpcHintResourceKey()],
+                                        SharedLocalizer[Model.Epc?.HeatingControls[1].HeatingControlEnumToEpcHintResourceKey()],
+                                        SharedLocalizer[Model.Epc?.HeatingControls[2].HeatingControlEnumToEpcHintResourceKey()]],
```
jdgage, line 54: Feels like all of this could be made simpler - we've got almost exactly the same line repeated six times:
```
SharedLocalizer[Model.Epc?.HeatingControls[i].HeatingControlEnumToEpcHintResourceKey()]
```
Could you map over the list?

SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
jdgage: I think there's a lot of branches here and it's uncomfortable to have:
1. this much business logic in the view
2. exceptions thrown in the view

Could you move more of this to a helper function?

jdgage (changes requested):

SeaPublicWebsite.Data/Migrations/20250113163718_AddHeatingControlsField.cs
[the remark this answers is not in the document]
Glenn-Clarke: I manually changed this, I couldn't start the project due to an error in this old migration, and it seems to be related to the way we declare the default value, as by removing it the issue went away.
The issue has only cropped up since the dependency upgrades, and I can't find which might have caused it. It looks like ef core only went up by minor versions.

I can't find what might be causing it, but I had a short look and I think that "{}" used to be understood to mean an empty int array, but now it isn't. 
"ARRAY[]::integer[]" seems to be a more explicit version of what it was previously doing (creating an empty int array as the default value).


For context: EFCore doesn't pick up the default value for these fields automatically from the models, so we have to add it in manually to set it.

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
Glenn-Clarke: I've looked over the values Temi has put on there and compared them to the logic. There doesn't seem to be any changes required, the current logic should handle them properly.

I'll pop our logic on the ticket in a more readable way, so DESNZ can give it a sign-off.

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
Glenn-Clarke: Empty list will result in no hint being shown.

The assessor notes in this field if no heating controls are present ("No thermostatic control of room temperature"), and that case is picked up by none of the specific controls being spotted in this field, and shows a suggestion accordingly.

If this field is null, the assessor hasn't explicitly indicated that there are none. So I've opted for no hint shown. Do you think this case is worth sign off from DESNZ?

SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
Glenn-Clarke: I'll have a look over these, some of them seem to be double checks

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
jdgage: I think bundle it in with the rest of your description of the behaviour: want to have it written down so we know what's happening the first time there's a bug ticket for someone not getting a recommendation.

SeaPublicWebsite.Data/Migrations/20250113163718_AddHeatingControlsField.cs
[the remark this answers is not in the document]
jdgage: Is there a GitHub issue for this? Or any writing online? I'd expect there to be if what EF Core's doing it fundamentally broken.

SeaPublicWebsite.BusinessLogic/ExternalServices/EpbEpc/EpbEpcDto.cs
Glenn-Clarke: It doesn't appear to have any issues with this

SeaPublicWebsite/Views/EnergyEfficiency/HeatingControls.cshtml
samyou-softwire: I've moved most of this now to the controller

alas I couldn't get the string interpolation to work in the controller, it'd keep losing the HTML formatting. I think this solution is a good middleground

samyou-softwire:
noticed that when running locally the migrations weren't applying for me. looks like the migration was missing a designer file. I've regenerated the migration with the same layout and it now seems to work

jdgage (approved):
Looks good. Only thing playing on my mind is the EF issue: I'd like to get to a conclusion on that before we call this closed.

SeaPublicWebsite.Data/Migrations/20250113163718_AddHeatingControlsField.cs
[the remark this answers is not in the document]
Glenn-Clarke: I can't find anything sort of comparing these two side by side except for a stack overflow answer of how to define a default value for an array.

I've traced the EF Core issue back to the dependency upgrade for NPGSQL (but just before EF was updated), but I can't find any reference to this exactly beyond people saying they've [used "{}" in the past](https://github.com/rails/rails/issues/9127)
Perhaps it could be [this?](https://www.npgsql.org/doc/release-notes/8.0.html#ilistt-mapping-now-requires-a-generic-npgsqlparametert)

I'm still not 100% on why it stopped working, but I do think this is how it should have been done all along (the explicit conversion to integer[] avoiding the implicit conversion of "{}" to Int Array)

Would be happy to call to look into this further if you have the time?

PR #443, opened 2025-07-29
PC-1707: Update recommendations on user returning

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1707)

# Description

adds two new db columns for whether recommendations were updated since last visit & when they were last updated.

now rerequests recommendations when:
- navigating past the 'Check your answers page'. previously it would only rerequest if the users answers changed
- entering in a referral code. previously it would always used saved answers

when receiving new recommendations, has new logic on setting the recommendation action where it searches existing recommendations for one with the same ID and same values (cost, savings etc). if it finds one it'll remember the action the user put for it.

also tracks whether recommendations were updated for some conditional text on the 'Your recommendations' page. the logic is as follows:

- if no changes to existing answers
  - set to false
- if changes to existing answers
  - were there any answers already saved
    - set to true
  - else no answers already saved (true if completing form for the first time OR changing answers, which already discards existing answers)
    - set to false

the extra logic on there being changes ensures users are not shown 'Your recommendations have changed' when submitting form for first time or changing their answers

when entering a code & there were changes to answers, send instead to the recommendations page instead of action page which will show the explanatory "Your recommendations have changed".

includes page structure changes identified in [PC-1747](https://beisdigital.atlassian.net/browse/PC-1747)

pending translations from DESNZ

# Checklist

- [x] If I have made any changes to content strings or resource files, I have followed [the documentation](https://github.com/UKGovernmentBEIS/beis-simple-energy-advice-beta/blob/main/docs/LocalisationDocumentation.md) (will do once translations are provided)
 
# Screenshots

<details>
your recommendations english
your recommendations welsh
action plan english
action plan welsh
</details>


SeaPublicWebsite/Services/EnergyEfficiency/PropertyDataService.cs
```diff
@@ -13,40 +15,91 @@ public class PropertyDataService(
...
+        if (existingRecommendations.Count != newRecommendations.Count)
+        {
+            return true;
+        }
+
+        return existingRecommendations.Any(er =>
+            !newRecommendations.Any(nr => nr.HasSameImpactAs(er)));
+    }
+
+    private void MergeExistingRecommendationActions(List<PropertyRecommendation> existingRecommendations,
```
samyou-softwire, line 90: happy to hear different ways on structuring this. ideally would not mutate at all but thought this could be tricky to structure

XakSenSoftwire (dismissed):
LGTM 👍 Left two questions for my own understanding

SeaPublicWebsite/Services/EnergyEfficiency/PropertyDataService.cs
samyou-softwire: left as blank. in this case it'd be that we've now received a new recommendation so they need to decide on it

LeoCie: I was going to suggest a change here to make this function a more-usual merge, than an update of the new recs with the old recs (e.g. return a list, not update the new recs). But am I correct in thinking that we want to have the new titles / summaries from the new recs always, and not take the old recommendations as they are in any case? In which case I think this does work -
I think lets rename this to something like "updateRecommendationsWithPriorActions" or similar (I can't seem to get the character count down there any more 😅). It makes it clear this updates the list

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1316,12 +1316,24 @@ await googleAnalyticsService.SendEvent(new GaRequestBody
...
 
-        var recommendations = propertyData.PropertyRecommendations;
+        // Always request the latest recommendations
+        var updatedPropertyData =
+            await propertyDataService.UpdatePropertyDataWithRecommendations(reference);
+
+        var recommendations = updatedPropertyData.PropertyRecommendations;
+
+        // On any change we show the user this page which explains why their recommendations may have changed
+        if (propertyData.RecommendationsUpdatedSinceLastVisit)
```
LeoCie, line 1326: is this propertyData up to date with the changes from the call to UpdatePropertyDataWithRecommendations?

It looks like we fetch `propertyData` in line 1312, but then save the update in recs to `updatedPropertyData`. Is this looking at the old version of `propertyData` intentionally?

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
```diff
@@ -1384,7 +1396,8 @@ public async Task<IActionResult> YourRecommendations_Get(string reference)
             NumberOfPropertyRecommendations = propertyData.PropertyRecommendations.Count,
             HasEmailAddress = false,
             BackLink = GetBackUrl(QuestionFlowStep.YourRecommendations, propertyData),
-            EnergyPriceCapInfo = propertyData.GetEnergyPriceCapInfo()
+            EnergyPriceCapInfo = propertyData.GetEnergyPriceCapInfo(),
+            RecommendationsUpdatedSinceLastVisit = propertyData.RecommendationsUpdatedSinceLastVisit
```
LeoCie, line 1400: Similar question here

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
samyou-softwire: ah good spot and this should reference `updatedPropertyData`. it does work as this (I suppose EF magic is updating `propertyData`?) though more clear to use the updated variable

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
samyou-softwire: this one should be fine since we update the properyData object and write it back out to the db when you put the code in, so we can retrieve it here

as in, `UpdatePropertyDataWithRecommendations` ends with
```cs
        await propertyDataStore.SavePropertyDataAsync(propertyData);
        return propertyData;
```

SeaPublicWebsite/Controllers/EnergyEfficiencyController.cs
LeoCie: Thanks, thought as much re. EF - probably some magic behind the scenes model caching makes it stay up to date somehow.

SeaPublicWebsite/Services/EnergyEfficiency/PropertyDataService.cs
samyou-softwire: have renamed this and squashed. needs approval again

PR #459, opened 2025-08-26
PC-2016: Pin google chrome version

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-2016)

# Description

to 139.0.7258.138-1

restructures install code a little to allow for this. we can't use version= syntax as google remove old versions from the PPA repository very quickly so the install will fail as the old version cannot be found. downloads the .deb install file instead manually and installs that which seems to be more stable

# Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,16 @@
 FROM mcr.microsoft.com/dotnet/aspnet:8.0 AS base
 # Install Chrome
 RUN apt-get update && apt-get -f install && apt-get -y install wget gnupg2 apt-utils
-RUN wget -q -O - https://dl.google.com/linux/linux_signing_key.pub | apt-key add -
-RUN echo 'deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main' >> /etc/apt/sources.list
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN wget --no-verbose -O /tmp/chrome.deb https://dl.google.com/linux/chrome/deb/
   && rm /tmp/chrome.deb
 
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends --allow-downgrades fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst fonts-freefont-ttf \
+    && apt-get install -y --no-install-recommends --allow-downgrades fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst fonts-freefont-ttf
```
XakSenSoftwire, line 12: Do we need `--allow-downgrades`? Is this what allows us to pin the google chrome version?

XakSenSoftwire (approved):
LGTM 👍 Left one question for clarity but happy to approve!

XakSenSoftwire (approved):
LGTM 🔥

